PR #18, opened 2020-03-20
Adds rudimentary social-media image, in Instagram crop.

4kbt:
Source image for surgical mask is CC1, Wikipedia

Source image for N95 is own work, CC0.


4kbt:
Changes complete.

PR #20, opened 2020-03-20
IA hierarchy changes

ctavispost:
Addresses issues 10 & 13

- Put states (h2), cities (h3), and donation site (h4) names in headers to increase readability, IA, and semantic accessibility
- Add CTA with hook to get people to read more (this is in a section that can easily be deleted if it is not desired); CTA links to donation info section
- put content in article container so that text max-width is 66ch at 1rem (page will not get too wide because of larger text)
- Center images and tweets to make them stand out and give the eye a bit of a bounce
- Put a little bit more space in between headlines to make reading easier

What it looks like: https://ctavispost.github.io/findthemasks/


r-pop:
Feedback: 
- headers for site names look great
- something went wrong with a link tag Nevada-Oregon
- I don't want the CTA right now, per previous comment not worried about actions below the fold. it's a quick read and I don't think we're losing folks. please remove text and button.
- because the tweets are part of the argument/narrative, I don't want eye bounce. please left align again.
- ok to center the mask images if you think it looks better - will defer to your judgement.

ctavispost:
Should be all set to go.

PR #34, opened 2020-03-21
Update index.html

erok415:
Editing formating and adding classes to match with styles


r-pop:
We are in the middle of a transition to automatic HTML generation based of a gsheet. We are not able to switch the listings structure at this time.

PR #35, opened 2020-03-21
Update style.css

erok415:


r-pop:
In general, I need a thorough description of changes especially from new contributors. Right now is a bad time for style changes as we're in the middle of some merges of manually written html.

Issue #2, opened 2020-03-19
Filter functionality for donation lists


r-pop:
Users should be able to filter by state, and materials accepted. (N95, surgical, booties, goggles, etc) and (opened box, unopened box).

eZanmoto:
I've implemented this in https://github.com/r-pop/findthemasks/pull/46, please let me know what you think.

jianshen:
Ooh @eZanmoto your filters look great! @r-pop feel free to ignore my PR. Great job putting all this together btw.

eZanmoto:
Thanks!

PR #44, opened 2020-03-21
Simple state nav filter for when this list gets longer.

jianshen:
Partially address Issue #2 so visitors can shortcut to/bookmark a particular state.


jianshen:
No longer needed.

PR #45, opened 2020-03-21
Fix formatting in `index.html`

eZanmoto:
This will allow styling to be applied to the site more easily.


graemer:
Outdated following cutover to automated index. Closing.

PR #62, opened 2020-03-21
Fix responsivity for mobile

ctavispost:
Removed `display: grid` from css classes. This should fix responsivity across browsers and screen widths.

–unsure *why* this was broken, but that's ok for now–


graemer:
Closing PR as this was prior to our cut-over to automated index.

PR #84, opened 2020-03-22
Versions of index.html and locations-list.js with a prototype map of …

susanashlock:
…dropoff locations added. 

Note bene / issues:
* API key is not checked in, so map will not load unless API key is added.
* Changes added in new files index-map.html and locations-list-map.js, since these changes aren't ready to go live.
* Note that only 4 locations are added to the map to limit number of geocoding calls to Maps API. I was running into rate-limiting errors when geocoding all 100+ locations.  I believe we can work around this by geocoding dropoff locations at time of import-from-spreadsheet.  (But I am unable to figure out how to run the import script myself, which is blocking me from adding this logic.)
* I ran into problems when I tried to add the map to the part of the page with the list of dropoff locations, so I left the map at the top of the page.
* Formatting of InfoWindows could be greatly improved.
* When you click on a second marker to bring up another InfoWindow, the first persists.  Not sure how to fix this. (Didn't investigate).

### DIFF
--- a/index-map.html
+++ b/index-map.html
@@ -1,0 +1,194 @@
+<!DOCTYPE html>
+<html>
+<head>
+    <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-161442455-1"></script>    
+    <script>
+  window.dataLayer = window.dataLayer || [];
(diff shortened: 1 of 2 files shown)


graemer:
This is great, thank you. Since resources are separate, merging now so we can all test / iterate more easily.

PR #87, opened 2020-03-22
Implement simple SSR using `envsubst`

eZanmoto:


graemer:
eZanmoto, could you add a bit more detail on the implications of this change. I’ve asked another engineer to take a closer look, but more context will help in the meantime.

eZanmoto:
Yes, sorry, the idea behind this is that `scripts/regenerate_index.sh` can be called every 5 minutes in a cron job to generate a single `ssr-index.html` file that would contain data and scripts. This is to overcome the various caching and rendering issues we've been facing. The proposal with this PR is that it won't replace `index.html` for now, but will be used for testing, and once `ssr-index.html` is deemed to be stable enough (not showing up blank or with old data) then we can update the script to replace `index.html` with the regenerator.

awong-dev:
This is an interesting fallback, but I think we're gonna be sticking with static files for the time being. Closing for now. Thank you!

PR #88, opened 2020-03-22
Make public directory that can be deployed from.

awong-dev:
Add a symlinks from the /public directory to the actual files
as a staging point for moving the files into /public.

This will make it easy to do a clean deploy of the client-side
code as well as keep it separate from the firebase cloud function
stuff.


awong-dev:
TBR @r-pop @graemer

PR #122, opened 2020-03-23
Added dynamic zoom to the map. 

spitznagel:
The map now zooms to the user's location and encompasses the three closest hospitals to the user. This zoom should scale in and out automatically based on the distance between hospitals and the user.

*** Note especially the addition of the geometry library in the initial API maps call on index-map.html. That library is required for the new function to work properly:
```JS
<script async defer
      src="https://maps.googleapis.com/maps/api/js?key=YOUR_KEY_HERE&libraries=geometry">
</script>
```

### DIFF
--- a/locations-list-map.js
+++ b/locations-list-map.js
@@ -349,7 +349,7 @@ function centerMapToNearestMarkers(map, markers) {
             }
 
             // sort the distances and set bounds to closest three
-            distances.sort((a, b) => a - b)
+            distances.sort((a, b) => a - b);
 
 
             // center the map on the user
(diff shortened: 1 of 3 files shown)


locations-list-map.js
```diff
@@ -332,13 +315,73 @@ function onFilterChange(elem) {
...
+                bounds.extend(loc);
+            }
+            map.fitBounds(bounds);       // auto-zoom
+            map.panToBounds(bounds);     // auto-center
+        }, function () {
+            alert('Could not get user location');
+        });
+    } else {
+        // Browser doesn't support Geolocation
+        alert('Could not get user location');
```
krithin: Please remove these two alerts before merging.

spitznagel: Removed alerts per comment in latest pull request.

Issue #124, opened 2020-03-23
Link for people who can't find a hospital no longer makes sense


raindrift:
There's a link on the site: "If there's not a site near you, visit our friends at getmeppe.org to let them know what you have available."

The link points to GetUsPPE, which now just embeds findthemasks.

We could instead link to or embed the PPELink donation form: https://ppelink.wordpress.com/ppe-donations/

Our existing volunteers can then find donation sites for people who fill it out, and create hospital records for whatever hospital they find for them.

r-pop:
Let's do that. Great catch.

r-pop:
Link please, embed will push content too far down.

PR #125, opened 2020-03-23
Moved link for donations that can't find a site from getusppe to ppelink

raindrift:
This is for issue #124


raindrift:
Note: when I have a moment longer, I can move our form embed into a new page on findthemasks.com instead. But this should fix the thing where the current link doesn't make any sense.

Issue #77, opened 2020-03-22
Consider adding "home-sewn" option for hospitals


4kbt:
We're starting to see hospitals including home-sewn masks on the list of resources they can accept. Might be useful to give that some data-structure?

r-pop:
Let's not do this quite yet. Waiting for more expert advice on which (if any) homemade PPE are protective enough for frontline HCW.

r-pop:
If we find a good spot, we could link "looking to donate home-sewn masks?" to this resource: https://docs.google.com/document/d/12a5YO0Z9RpHZk9Zkzl4NOj9CbjzhFfoKjPLFFC-21LU/edit

cchang-adi:
A few hospitals on our list have explicitly called for homemade masks and even provided links to patterns. There at least should be a filter category for this option, or even an explicit form checkbox, since they seem to want it.

raindrift:
I think the right thing to do here is to link to maskforce.org, or maybe even directly to their intake form (https://airtable.com/shrCKIswWL0E5lrXk). They ask a lot of very specific questions, like what materials are acceptable, what the mask will be used for (to cover an n95?), and whether the hospital has a specific design they want used. I think it makes more sense for us to send people there than to try to capture all the data they need.

jenninoordhoek:
Airtable has an embed code and an API - problem with API is it will only take 5 calls per second, but hopefully that doesn't apply to the form.

r-pop:
As next step, we need a designer to think about how this locks up with the existing content. Perhaps a call-out with two diversion to other efforts:

Maker? Sewing masks? -> maskforce.org
Large inventory to donate? -> [appropriate resource that can do logistics/funding TBD from the intergroup, suggestions welcome]

r-pop:
@cjimmy maybe something to look at tomorrow

jenninoordhoek:
And then circle back to the same distribution? because sewn masks and donated masks from individuals need to be the same channels.

raindrift:
I spoke with Danielle, and she's on board for having our forms mutually link to one another. I'll go ahead and put together a PR for this that at least does the linking, pending some more pleasant design.

PR #127, opened 2020-03-23
Move intake form into an embed, add link to MaskForce

raindrift:
This addresses #77. I spoke with Danielle from Mask Force, and she agreed that because they request very specific information from hospitals, it makes sense for us to link to one another instead of attempting to take requests for home-sewn masks directly.


raindrift:
Ah damn. I totally rebased this PR onto the wrong thing. Going to open a new PR instead.

Issue #117, opened 2020-03-23
URL targets for specific states (potentially regions?)


patricknelson:
Find a method to ensure that we can have dedicated landing pages for states, or even regions (if feasible). 

Idea: Include query sting params to facilitate filtering via JS, including:

- `state`, accepting ISO 3166-1 alpha-2 code (should be robust, just normalize to lowercase)
- `accept`, filters accepted items via simple text search, maybe using fuzzy search, e.g. https://fusejs.io/ (but literal match might be perfectly fine as well).

patricknelson:
Currently getusppe.org has the following portal (noted in slack): https://getusppe.org/states/

findthemasks.com could potentially leverage a similar functionality at OUR level by pulling in filters using query string parameters. Both here and also in the embed noted in #113.

patricknelson:
**Future:** Would be awesome to find some way to also target narrow regions, e.g. SF Bay Area and/or Boston, etc. Ideally we'd probably end up using a radius search based on lat/long data (very feasible). Can this be performed client side?

patricknelson:
Fixed in #134

PR #134, opened 2020-03-23
Set the initial filter state for states using the URL query params.

krithin:
This works with both repeated params and comma-separated values for a single param.

### DIFF
--- a/locations-list-map.js
+++ b/locations-list-map.js
@@ -235,6 +235,15 @@ document.addEventListener("DOMContentLoaded", function() {
 
     const htmlSnippets = toHtmlSnippets(data_by_location, null);
     $(".locations-list").html(htmlSnippets.join(" "));
+
+    const searchParams = new URLSearchParams((new URL(window.location)).search);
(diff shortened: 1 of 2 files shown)


krithin:
This is the 'states' part of https://github.com/r-pop/findthemasks/issues/117 ; accepted items could be done with the same approach if that's useful.

krithin:
> Deployed assets are coming from /public, so please move this change to /public/locations-list-map.js

Done!

krithin:
You can test this by loading `public/index.html?state=al,co&state=wv`

patricknelson:
Awesome! Here's a live proof of concept now that it's live: 

 **Full site:** https://findthemasks.com/?state=al,fl

**iframe version:** https://findthemasks.com/give.html?state=al,fl

Thanks @krithin!  I'll close out #117

patricknelson:
@Shuhanhe1 this is now ready for implementation on https://getusppe.org/states/, see URL syntax above. ☝️

PR #135, opened 2020-03-23
Cleaning up "smart" quotes in meta tag attributes. And a shout out.

patricknelson:
Cleaning up "smart" quotes in meta tag attributes. And dropping in a quick shout out (I plan to do more).

**Please test** before merging; I'm pretty sure `<p">` was a mistake but shouldn't exist, but I am an alien from outer space, so I could be wrong. 👨‍🚀


patricknelson:
Closing in favor of #137

PR #137, opened 2020-03-23
Cleaning up "smart" quotes in meta tag attributes. 

patricknelson:
Second pass at original #135 (before The Great Reform).

----

Cleaning up "smart" quotes in meta tag attributes. And dropping in a quick shout out (I plan to do more).

**Please test** before merging; I'm pretty sure `<p">` was a mistake but shouldn't exist, but I am an alien from outer space, so I could be wrong. 👨‍🚀

### DIFF
--- a/public/whoweare.html
+++ b/public/whoweare.html
@@ -37,7 +37,7 @@
     <a href="https://www.linkedin.com/in/catherinelchang">Catherine Chang</a><br>
     Melanie Cohen<br>
     Melissa Kerr<br>
-    <a href="http://twitter.com/sean_m_kelleher">Seán Kelleher</a><br>
+    <a href="http://twitter.com/sean_m_kelleher">Se&aacute;n Kelleher</a><br>
     <a href="https://github.com/cjimmy">Jimmy Chion</a><br>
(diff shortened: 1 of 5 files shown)


public/whoweare.html
```diff
@@ -37,19 +37,21 @@ <h1>#findthemasks</h1>
     Catherine Chang<br>
     Melanie Cohen<br>
     Melissa Kerr<br>
-    <a href="http://twitter.com/sean_m_kelleher">Seán Kelleher</a><br>
+    <a href="http://twitter.com/sean_m_kelleher">SeÃ¡n Kelleher</a><br>
```
patricknelson: It's interesting this came out this way in Github, it looks like this in my local dev (unaffected). I think maybe I'll use `&aacute;` here to be safe since this looks like an old fashioned ISO-8859-1 encoding/transliteration bug (even though everything's defined as UTF-8). 🤔

patricknelson:
@graemer Just rebased and cleaned up my code; I removed the shout out 😞 That created a conflict, of sorts. Just hoping to help fix some encoding errors here 😄

Issue #140, opened 2020-03-23
www.findthemasks.com should redirect to findthemasks.com


patricknelson:
This should help reduce likelihood of duplicated content. Also ensure everyone is linking to the same URL's. The redirect should be a perm redirect (301) and should retain the request URI (i.e. if you go to https://www.findthemasks.com/give.html?state=ca it should redirect you to https://findthemasks.com/give.html?state=ca).

r-pop:
Hi, I don't think this is necessary. I have dreamhost set up to allow both www. and bare, and the content is the same. no duplication.

r-pop:
If you disagree please explain more thoroughly

patricknelson:
It's just a best practice to ensure only one domain is in use; see more here: https://support.google.com/webmasters/answer/66359?hl=en

Alternatively, you could "canonicalize" with the appropriate meta tags, see: https://moz.com/learn/seo/canonicalization, e.g. 

```html
<link rel="canonical" href="https://findthemasks.com/" />
```

Unfortunately that would need to be done for every single page (with the appropriate URL plugged in). Usually it's just easier to do the domain redirect 😄

patricknelson:
Closing; not really worthwhile, considering the testing/validation effort 😄

PR #141, opened 2020-03-23
Fix for #140: Redirect users away from 'www' prefix (hostname agnostic), carry over request URI.

patricknelson:
Fix for #140: Redirect users away from 'www' prefix (hostname agnostic), carry over request URI.

You can test this with `curl -v -H 'Host: www.findthemasks.com' http://127.0.0.1/` as well if you spoof the domain in the host header (e.g. see `-H` param). But you can also test locally after commenting out the HTTPS redirect and manually tweaking the redirect line and then using your hosts file to map both domains to your localhost machine (or whatever machine you're doing dev on, I used `dev.findthemasks.com` as my local dev domain). See below where I demonstrate this.

Also note: This also **fixes** a bug in the `.htaccess` file, since the first `https://` rewrite rule wasn't even working at all since `RewriteEngine On` must be called first. 

**⚠️Caution:⚠️** I'm not sure how those first two lines will work, it depends on the underlying architecture and if Apache is really convinced that `HTTPS` is `on`. Usually upstream servers (like yours) will be over plain text HTTP and therefore `HTTPS` will not actually be `on`, however CDN services like CloudFlare and Akamai will typically incorporate the `X-Forwarded-Proto: https` header, which you could check instead. So be sure to test this with the ability to revert quickly, or test on another domain.

### DIFF
--- a/public/.htaccess
+++ b/public/.htaccess
@@ -1,5 +1,13 @@
+RewriteEngine On
+
+# First: Always ensure we're going over HTTPS.
 RewriteCond %{HTTPS} !=on
 RewriteRule ^(.*) https://%{HTTP_HOST}%{REQUEST_URI} [L,R=301]
+
+# Redirect to the WWW domain after already on HTTPS.
+RewriteCond %{HTTPS} on
+RewriteCond %{HTTP_HOST} ^www\.(.*)$
+RewriteRule .* https://%1%{REQUEST_URI} [L,R=301]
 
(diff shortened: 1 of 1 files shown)


graemer:
@awong-dev - are you comfortable reviewing this change? I'm not familiar here.

patricknelson:
Closing; not really worthwhile, considering the testing/validation effort 😄

PR #145, opened 2020-03-23
Load the google maps script with 'defer' only, not 'async'.

krithin:
This should prevent a race condition where initMap (via
DOMContentLoaded) can sometimes be called before the map script has been
downloaded.


krithin:
Fixes https://github.com/r-pop/findthemasks/issues/143

graemer:
Any know risk to removing the async? @krithin

alex:
I don't think there are, from the Mozilla docs, this does exactly what we want.

PR #151, opened 2020-03-23
Move intake form into an embed, add link to MaskForce

raindrift:
This addresses #77. I spoke with Danielle from Mask Force, and she agreed that because they request very specific information from hospitals, it makes sense for us to link to one another instead of attempting to take requests for home-sewn masks directly.

I made a mess rebasing PR 124, so this a new PR with pretty much the same code.


raindrift:
Screenshot 2020-03-23 11 10 42

Here is a screenshot of the form with the embed in place.

graemer:
@raindrift sorry to be a bother. Would you mind splitting the request form and the index update into separate PRs? That way, we can merge and deploy request and do some on-device testing, and then update index once validated. Sound OK?

raindrift:
Done! I'll make a new PR for the index re-pointing

PR #153, opened 2020-03-23
Fix #149 add query params to hide map / filters

### LINKED ISSUE #149: Add ability for give.html to dynamically display filters & map via query-params
For external consumers, we need a few additional query param capabilities for the embed resource.

Specifically,

`?show-map=true/false` - drives visibility of map component; if false, the maps API should not be hit

`?show-filters=true/false` - drives visibility of filters. this will be used for per-state pages within partner sites, who will use the ?state=WY, etc. for filtering content

Maybe...

`?region=US/EU` - this will drive which JSON file gets pulled. We have an 11AM meeting with CERN where we may be standing up an EU version. This will allow us to switch the queried dataset via URL param.

nburt:


graemer:
@nburt Added a small styling commit. Merging momentarily.

PR #157, opened 2020-03-23
query param to hide list, zoom map to state

nburt:


nburt:
If we can to use the Geocoding API the zooming would be cleaner. To do so we'd need to look up the state name using an object like this: https://gist.github.com/mshafrir/2646763

We need the mappings because for that API searching for "State of New York" works but "State of NY" doesn't

public/locations-list-map.js
```diff
@@ -344,8 +350,92 @@ function onFilterChange(elem) {
...
+  "WY": { "lat": 43.075968, "lng": -107.29028 }
+};
+
+function centerMapToState(map, state) {
+  const stateLocationMapping = stateLocationMappings[state.toUpperCase()];
+
+  // below code left for future use of geocoder API which would be more accurate in zooming
+
+  // const geocoder = new google.maps.Geocoder();
+  // geocoder.geocode({'address': `State of ${stateMappings[state]}`}, (results, state) => {
```
nburt: if we do use the geocoding api we should update the address to `${stateMappings[state.toUpperCase()]}, USA`

graemer:
@nburt - This is great. I'll pull down and do some local testing. Thank you.

Issue #160, opened 2020-03-24
Add i18n


raindrift:
Make all the strings translatable

raindrift:
Working on this now.

patricknelson:
🐿 Shipped! Can close.

volantwish:
🎉

PR #164, opened 2020-03-24
Added i18n.js, extracted strings from key files

raindrift:
Here's basic i18n, for #160. Translations are stored in i18n.js. Normally this would be standalone json, but loading that breaks in our local-file based dev workflow. If we decide we want to use a tool like translatewiki, it should be straightforward to add translations as json for languages we don't use in dev.

So far I've only internationalized locations-list-map.js and give.html, and there is no language selector control (just defaults to browser locale)


public/locations-list-map.js
```diff
@@ -167,11 +167,12 @@ function toHtmlSnippets(data_by_location, filters) {
     for (const city of Object.keys(cities).sort()) {
       const entryLines = [];
       for (const entry of cities[city]) {
-        const name = entry["What is the name of the hospital or clinic?"];
-        const address = entry["Street address for dropoffs?"];
-        const instructions = entry["Drop off instructions, eg curbside procedure or mailing address ATTN: instructions:"];
-        const accepting = entry["What are they accepting?"];
-        const will_they_accept = entry["Will they accept open boxes/bags?"];
+
+        const name = entry[$.i18n("ftm-sheet-hospital-name")];
```
krithin: I believe these five strings are actually embedded in the header row of google sheet that underlies the site, and I'm not sure if they should be translated - flagging for @graemer to take a look.

awong-dev: Yes. don't touch these right now. I'm going to fix them so they're not localized in a bit.

raindrift:
They are the google sheet header names, as it says in the comment next to
them.

I localized them because I expect that people using the site in other
languages will also translate the google form, and then their headers will
be different.

-Ian

public/locations-list-map.js
raindrift: I did this intentionally, but on second thought I'll move them to a separate object so that their purpose is more clear.

raindrift: Following up: I added strings for these because they will be translated for anyone who translates the form. But I agree that just translating them like other locale strings is a mistake, since they need to match the sheet exactly. The comment I had in there was clearly not enough to call them out because y'all didn't see it. :)

I moved them into their own separate object.

raindrift:
Aha! I see someone fixed the column names so they're no longer big natural-language strings. That is way better! I've updated this PR in accordance.

public/give.html
```diff
@@ -140,36 +145,46 @@
...
+
     <meta property="og:image" content="https://findthemasks.com/images/n95.jpg"/>
 </head>
 
 <body>
 <main class="content">
     <div class="container" style="padding-bottom: 40px;">
       <div class="map-container" style="display: none;">
-        <h2 style="padding: 30px 0px 20px 0px;">Map of donation sites</h2>
+        <h2 style="padding: 30px 0px 20px 0px;" class='i18n' data-i18n="ftm-map-of-sites">Map of sites™</h2>
```
krithin: Nit: stray ™

public/give.html
```diff
@@ -140,36 +145,46 @@
...
         <div id="map"></div>
       </div>
 
       <div class="locations-container" style="display: none;">
-        <h2 style="padding: 30px 0px 5px 0px; margin: 0px;">List of donation sites</h2>
-        <p style="padding-bottom: 10px">To add a donation site, update information, or request we take down information, please
-          <a href="https://docs.google.com/forms/d/e/1FAIpQLSfgCpK5coPVFC6rJrE7ZhimiZuDoEaL6fo6gYqxsN_FIpJZhg/viewform">fill in this form</a>.</p>
+        <h2 style="padding: 30px 0px 5px 0px; margin: 0px;" class='i18n' data-i18n="ftm-list-of-sites">List of donation sites</h2>
+        <p style="padding-bottom: 10px"> <span class='i18n' data-i18n="ftm-request-update">To add a donation site, update information, or request we take down information, please</span>
```
nburt: Mostly out of curiosity. Does this library handle tags at all? Is it possible to wrap the entire sentence to have the full context rather than it being split into 2 strings?

raindrift:
It does. You need to change the data tag to indicate that it’s allowed, and
then the string can include markup.

Issue #66, opened 2020-03-21
Site reportedly doesn't render on Microsoft Edge


rezendi:
Report from a nurse friend in Maine. Not sure which version at the moment - have enquired.

rezendi:
Update: "Microsoft edge 42.17134.1098.0" "Microsoft edge HTML 17.17134"

r-pop:
If someone has Windows and can repro, please let us know here.

r-pop:
@marcpopmsft does this repro for you? This is EdgeHTML Edge, not Chromium Edge, right?

cchang-adi:
Repro'd on Edge on Windows 10. Filters don't work - clicking doesn't hide other states information.

volantwish:
On Edge, I'm not seeing the map, filters, or list display at all...the error in the console is `Object doesn't support property or method 'flat'`, pointing at [this line](https://github.com/r-pop/findthemasks/blob/a83cdc635af6cf3246868ac0302760455530d6d0/public/locations-list-map.js#L236). We're going to need to replace the arrow functions throughout the JS for IE/Edge support. If anyone's picking this up, please comment, otherwise, I can take a look at this tomorrow.

volantwish:
Oops, was looking at the wrong line. I've got a fix for Edge. There's still more work to do for older IE, though.

vdalal:
Site renders fine with:
* Microsoft Edge 80.0.xyz

r-pop:
Great! I think this was for EdgeHTML Edge, and volantwish fixed it.

PR #165, opened 2020-03-24
Add Edge support

volantwish:
Edge support for #66. There's some followup work to do to make the site work for older browsers. (But I'm not set up to test those right now.) Might make most sense to add a new bug for older IE versions.


patricknelson:
Confirmed that this is working in Edge on my machine (was broken before this). Great work!

image

PR #169, opened 2020-03-24
Scroll the list of locations into view when the filter state changes.

krithin:
But don't autoscroll there on page load so users get to see the top of
the page first.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -215,7 +215,7 @@
           <form name="filters" class="filters-list"></form>
         </div>
         <div class="locations-container" style="display: none;">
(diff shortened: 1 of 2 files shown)


krithin:
This addresses https://github.com/r-pop/findthemasks/issues/71 but I have some reservations about whether it's a good idea - in particular it makes selecting multiple states with the UI annoying because it scrolls away each time. It's fine if we think making single-state selection easy is the more important use case though.

drwestco:
Yeah, this behavior is kind of obnoxious, especially when trying to select multiple items from the states and/or accepted items filters.

PR #175, opened 2020-03-24
Use a stable header row

awong-dev:
This should free us up to rename fields in the form and localize the sucker.


krithin:
Did you mean to include the CSS change in this PR?

awong-dev:
It's just a whitespace fix, but I'll break it out. one sec.

awong-dev:
PTAL

awong-dev:
thank you!

PR #178, opened 2020-03-24
Revert "Check in @susanashlock's geocoder script that was in the google sheets app."

krithin:
Reverts r-pop/findthemasks#174

It turns out that this doesn't work when remotely loaded into the google apps script environment.


susanashlock:
Actually, this does appear to work!  Thanks, Krithin@!

PR #182, opened 2020-03-24
added basic stats page

nburt:
extracted shared function

Could someone with a little more knowledge of browser compatibility confirm that the type="module" will work here so we can share that function? I haven't written js without babel in a long time...

Once i18n.js is back in a can merge that and add in i18n strings


awong-dev:
It's okay to assume es2015+ and modern browser.  The rest of the site already does and most of traffic is mobile.

PR #194, opened 2020-03-25
Geocode blank rows at data.json generation time

krithin:
This PR doesn't include any mechanism to cache the geocode result, so we should take care to check that it won't burn through our maps API quota too fast.


krithin:
Possibly addresses https://github.com/r-pop/findthemasks/issues/144

graemer:
@krithin @awong-dev @alex @susanashlock - I'm not comfortable reviewing this one properly. Can someone (or multiple) sign up to go through this in detail?

krithin:
> @krithin @awong-dev @alex @susanashlock - I'm not comfortable reviewing this one properly. Can someone (or multiple) sign up to go through this in detail?

Careful review is especially appreciated because we were working with a different file in local development (to avoid all the firebase auth stuff) and then I merged the changes into index.js by hand and haven't been able to run it locally to verify that the result of that merge works.

functions/index.js
```diff
@@ -116,19 +118,66 @@ async function snapshotData(filename, html_snippet_filename) {
...
+  return client
+  .geocode({
+    params: {
+      address: address,
+      key: GOOGLE_MAPS_API_KEY,
+    },
+    timeout: 1000 // milliseconds
+  })
+  .then(r => {
+    console.log(r.data);
```
awong-dev: This is going to log spam us.  Do we really need it?

functions/index.js
```diff
@@ -116,19 +118,66 @@ async function snapshotData(filename, html_snippet_filename) {
   return [data, html_snippets];
 }
 
+// Fetch lat & lng for the given address by making a call to the Google Maps API.
+// Returns a promise.
+async function getLatLng(address, client) {
+  return client
```
awong-dev: nit...you're using a promise in an async function. I think you could have just done
const r = await client.geocode(...);
if (r.data.results & r.data.results.length > 0) {
  ....
}

functions/index.js
```diff
@@ -142,7 +191,7 @@ function toDataByLocation(data) {
     const entry_obj = {};
     headers.forEach( (value, index) => {
       if (entry[index] !== undefined) {
-        entry_obj[value] = entry[index].trim()
+        entry_obj[value] = (typeof entry[index]) === 'string' ? entry[index].trim() : entry[index];
```
awong-dev: when is this even not a string??

functions/index.js
```diff
@@ -116,19 +118,66 @@ async function snapshotData(filename, html_snippet_filename) {
...
+    // Do geocoding and add lat/lng to data iff we don't already have a latLng for
+    // the address *and* the address has been moderated.  (In this code block, all
+    // data has already been moderated.)
+    const address = entry[addressIndex];
+
+    // Check if entry's length is too short to possibly have a latitude, we need
+    // to geocode.  If the value for lat is "", we also need to geocode.
+    if ((entry.length < (latIndex + 1)) || (entry[latIndex] == "")) {
+      try {
+        let lat_lng = await getLatLng(address, maps_client);
```
awong-dev: nit: const. You're never reassigning the variable.

functions/index.js
```diff
@@ -116,19 +118,66 @@ async function snapshotData(filename, html_snippet_filename) {
...
+        entry[latIndex] = lat_lng.lat;
+        entry[lngIndex] = lat_lng.lng;
+        console.log(entry);
```
awong-dev: remove log?

functions/index.js
```diff
@@ -116,19 +118,66 @@ async function snapshotData(filename, html_snippet_filename) {
...
+      } catch (e) {
+        console.error(e);
+        entry[latIndex] = 'N/A';
```
awong-dev: should this be empty-string? The downstream code doesn't have to check against a sentinel value.

drwestco:
> This PR doesn't include any mechanism to cache the geocode result, so we should take care to check that it won't burn through our maps API quota too fast.

I think that's important before this is accepted. Unclear how often the data.json is generated, and how many entries get re-geocoded on average.

awong-dev:
@drwestco yes. Agreed. The data.json is triggered by a manual process so it's frequently, but not machine-slamming frequent.

I'm working on a write-back solution currently which should solve this.

functions/index.js
krithin: Done.

functions/index.js
krithin: Done. Removed.

functions/index.js
krithin: Done.

functions/index.js
krithin: Done. Sorry I forgot to do a pass to remove all of them.

functions/index.js
krithin: I think 'N/A' is consistent with what's currently in the spreadsheet?

functions/index.js
krithin: For a column containing a lat/lng value - it has numeric type in that case, so calling trim() on it fails.

awong-dev:
Oh...I just realized we put the geocode in the wrong spot.  But let's merge this and I can fix.

functions/index.js
awong-dev: ok

PR #199, opened 2020-03-25
Backfill geolocation into spreadsheet

### LINKED ISSUE #144: Geocoding breaks daily
When I run the geocoding script (Apps Script, accessible from Tools > Script Editor in Google Sheets), I now get this error:

Exception: Service invoked too many times for one day: geocode. (line 38, file "Code")

According to this documentation, it might not be possible to work around in Apps Script: https://developers.google.com/apps-script/reference/maps/maps#setAuthentication(String,String)

Relevant documentation pasted below:
------
Enables the use of an externally established Google Maps APIs Premium Plan account, to leverage additional quota allowances. Your client ID and signing key can be obtained from the Google Enterprise Support Portal. Set these values to null to go back to using the default quota allowances.

Note: This method doesn't work with API keys. Additionally, please note that Premium Plan is no longer available for new customers. If you don't already have a Premium Plan license, please don't call setAuthentication(clientId, signingKey). You are able to use the Maps methods with the default quota allowances.
---------
If we can't figure out how to get quota to run from Apps Script, an alternative would be to add this logic to functions/index.js so that we do the geocoding using an API key at spreadsheet-read time.  The problem with this is that since the data.json is recreated from scratch every time, we'd be making far more geocoding calls than we need because we'd be re-geocoding data that we already have.  e.g. if we assume that we add 200 hospitals in a day, and re-run geocoding every time we add a hospital, that's ~20,000 geocoding calls / day.

I'll explore both option 1 (apps script) and credits for option 2 (updating index.js).

awong-dev:
Mostly fixes #144 .

Currently it only fills 5 lat/longs at a time. There needs to be rate limiting code to stay under the 50qps of the Geolocation API, otherwise the geolocation server starts rejecting requests.

### DIFF
--- a/functions/index.js
+++ b/functions/index.js
@@ -15,11 +15,19 @@ const CONFIG_CLIENT_SECRET = functions.config().googleapi.client_secret;
 const CONFIG_SHEET_ID = functions.config().googleapi.sheet_id;
 const GOOGLE_MAPS_API_KEY = functions.config().findthemasks.geocode_key;
 
+const COLUMNS = [
+'A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M',
+'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z',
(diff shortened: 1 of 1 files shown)


awong-dev:
It's deployed so I'm committing.

TBR: @susanashlock @krithin

Issue #166, opened 2020-03-24
Support older browsers


volantwish:
Followup to #66  to add support for older browsers (https://caniuse.com/#feat=arrow-functions)

patricknelson:
Template literals as well, if supporting IE11: https://caniuse.com/#feat=template-literals 

Would be great to have analytics stats on this. I know on our websites we typically drop support for browsers that are at or below 2%-3% market share (just to save on dev resources).

vdalal:
Related to #66
Site renders fine with:
* Microsoft Edge 80.0.xyz/Windows 10

vdalal:
Filters:
* "Zip Code": works
* "City, State": works
* "Use my location": does not work
* "Use my location": option disappears after one clicks on it
* "State": works
* "By Accepted Items": works

Tested with Microsoft Edge 80.0.xyz/Windows 10

r-pop:
Keep in mind that IE and older Edge will be overrepped in healthcare. I have seen Win95 in hospitals when there as a patient, even in tech-centric Seattle.

saikatbhadra:
Putting the same caveat as @r-pop - Healthcare workers will be on more archaic systems. However, as long as they can go to to the site and fill out the form their main use case will be handled.

saikatbhadra:
Here are the browser stats. I was noticing some patterns changing (for example desktop usage going up) so please access GA for the latest stats:

Chrome version stats:
![Screen Shot 2020-03-31 at 4 22 09 PM](https://user-images.githubusercontent.com/2250961/78084018-f954ca80-736b-11ea-99f7-5f1881bc75ec.png)

Safari stats (what is 604???):
![Screen Shot 2020-03-31 at 4 24 30 PM](https://user-images.githubusercontent.com/2250961/78084090-2ef9b380-736c-11ea-9b6f-5ac11e6adba2.png)

Firefox stats:
![Screen Shot 2020-03-31 at 4 26 25 PM](https://user-images.githubusercontent.com/2250961/78084192-65cfc980-736c-11ea-9879-497ac244740a.png)

saikatbhadra:
Edge stats:

![Screen Shot 2020-03-31 at 4 29 45 PM](https://user-images.githubusercontent.com/2250961/78084376-e1317b00-736c-11ea-8341-bf93c4d92ac5.png)

IE stats:
[...]

PR #211, opened 2020-03-26
Fix bug when filtering by state in Edge.

patricknelson:
This adds a polyfill for Edge to fix bug when filtering by state; due to lack of support for HTMLCollection.prototype.forEach. See below.

### DIFF
--- a/public/locations-list-map.js
+++ b/public/locations-list-map.js
@@ -289,6 +289,13 @@ window.onFilterChange = function (elem, scrollNeeded) {
   }
 };
 
+
+// Polyfill required for Edge for the .forEach methods above, since this method doesn't exist in that browser.
+if (window.HTMLCollection && !HTMLCollection.prototype.forEach) {
+  HTMLCollection.prototype.forEach = Array.prototype.forEach;
+}
+
+
 const stateLocationMappings = {
   "AK": { "lat": 63.588753, "lng": -154.493062, zoom: 3.5 },
   "AL": { "lat": 32.318231, "lng": -86.902298, zoom: 6 },


akainic:
Hey @patricknelson! Want to confirm if this fix is necessary? According to [caniuse](https://caniuse.com/#search=forEach), Edge does support the use of `.forEach`. Curious why you seem to have run into an issue with it. Was it an older version, or perhaps IE?

Separately, I'm not sure if adding polyfills for specific JS methods is the best approach. It could get messy fast (if there are more methods in the future we want to support, we'd have a long list of polyfills in our code!), and there are good, if not better, alternatives for `.forEach` that we could easily use instead (e.g., plain old `for` loop or `for... of`).

Let me know if this is still creating an issue for you thought, happy to dig into this further!

patricknelson:
The problem was not `.forEach` missing on `Array` or `NodeList`, it was missing on `HTMLCollection` (screenshot in the first comment). This is Edge 17.

Again, I agree but this is a quick fix. It is messy and I wouldn't have used .foreEach() but there are also arrow functions and JS templates as well as modules. Right now we're optimized for speed and not entirely cleanliness. I'd suggest a refactor too; 

In fact, there's a ticket for that; would you like to submit a PR for #166?

awong-dev:
Yikes. This is indeed ugly, but if it patches Edge for now, let's do it and submit a follow-up to clean up.

The big question is how much we care about older browsers. If we do actually care, we should create a real deploy pipeline with babel and webpack or something.

patricknelson:
Precisely my thought @awong-dev. That requires a more comprehensive pipeline and dev workflow which would be good to dig into eventually.

Issue #208, opened 2020-03-25
Map loading slowly - add progress indicator


r-pop:
we've reached a point where the map takes a noticable second to display on pageload. Can we as a stopgap put a "Loading..." indicator in its place

The text "Loading..." is fine

r-pop:
This is what it looks like while loading. It looks like there is no data. We need to say "Loading list of donation sites..."
![slowload](https://user-images.githubusercontent.com/11252938/77581197-81356300-6e9a-11ea-8748-bfce960af80e.png)

Panometric:
A few suggestions:
1. At 1400+ datapoints, it is never going to render quickly. How about using the attached image as a place holder until it renders?  We could also make this image fuzzy. 
![image](https://user-images.githubusercontent.com/716218/77599021-a5a43600-6ec0-11ea-99cc-5b2ff31988b2.png)

2. There is a usage limit 500 requests per second (QPS). If the site is too busy, the requests will just fail. So maybe put the actual map under a search button?
https://developers.google.com/maps/documentation/embed/usage-and-billing

patricknelson:
After further analysis, it appears a majority of the latency appears to be in network overhead. Specifically, the downloading, decompressing and parsing the https://findthemasks.com/data.json file (page is able to runder while XHR kicks off in another thread to pull that file down) which is why we still see the page waiting for it to DL and doesn't seem to have anything to do with the map.

Turns out there may not be a dramatic difference with vs. without Google Maps, see below (alternating between https://findthemasks.com/give.html vs. https://findthemasks.com/give.html?hide-map=true. In this case, if `hide-map=true` it completely avoids initializing Google Maps and had similar delay. 

Still digging into this.... see below, though.

![2020-03-25_17-54-25](https://user-images.githubusercontent.com/4269377/77599647-a211ae80-6ec2-11ea-8c18-97e773a8b7ee.gif)

patricknelson:
Working on wrapping this up.

patricknelson:
Ok fix here: #216

PR #216, opened 2020-03-26
Fix for #208: Adding loading animation/spinner while data.json is loaded and parsed.

patricknelson:
Fix for #208: Adding loading animation/spinner while data.json is loaded and parsed.

Preview below. Using https://feathericons.com/, credit w/ license info in `README.md`.


patricknelson:
Oh @r-pop possible regression here -- note that this **starts out** with the "List of donation sites" heading. Let me know if that's ok! IMHO, that is an improvement. It's an artifact of aligning BOTH the `index.html` and `give.html` file structures.

cc @graemer @awong-chromium @awong-dev

public/give.html
```diff
@@ -157,6 +157,32 @@
         #map-stats, #list-stats {
           font-weight: 300;
         }
+
```
akainic: Hey @patricknelson! Just diving in here so definitely don't have all the context, but wondering why we're including the CSS in a `style` tag instead of including a link to the css file? It might remove the need to copy the same set of styles into both html files.

public/locations-list-map.js
```diff
@@ -1,6 +1,7 @@
 import toDataByLocation from './toDataByLocation.js';
 
 function createFiltersListHTML() {
+  $('.locations-loading').hide();
```
akainic: At this point it's possible that the map will still not have loaded. This function `createFiltersListHTML` is just about creating the state filters and not related to the map loading.

I think there's something you can do where you add an event listener on the `tilesloaded` event from Google Maps, which should tell you more accurately when the map has loaded (see this [stackoverflow post](https://stackoverflow.com/questions/33234945/how-to-detect-if-google-map-loaded-successfully)). I was able to reproduce this in the browser console, so I think it should work!

public/give.html
patricknelson: I don't like it either, it's the convention they went with. Not 100% sure, but @graemer  may have more perspective on that; reach out to them on our slack instance for more context.

And yeah, I agree. Right now it's very redundant and prone to errors (hence the need in this commit to restructure the HTML for consistency)!

public/locations-list-map.js
patricknelson: That's true, but there are a few reasons why I went this way (and keep in mind that I'm still conforming to the exiting JavaScript structure).

The JS is structured in a certain way the number one important goal here is to ensure it's as intuitive as possible (i.e. hide one thing and show the other, they're contextually relevant albeit it's not 100% accurate in terms of actual execution vs. intent). In this case placing it here for the most obvious place to find it, helps adhere to [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).


Given this site is subject to VERY rapid change, I'm conforming to the existing structure. If I built it I may restructure it and move **both** the `hide()`/`show()` methods OUT of the `createFiltersListHTML()` function because the name _implies_ that it's only job is to "create HTML", but that's not all it does. So, in the interest of speed I just rolled w/ what's already here.

Make sense?

patricknelson: Here you go, check this out  @akainic: 9c1db55 Thanks for the guidance :)

drwestco:
Unless hide-map=true, we know we're going to show the map, so may as well have the header + map container visible right away. The loading spinner can go in the map container until the map is ready to replace it.

patricknelson:
I think we're pretty much set and this is ready to ship :shipit:.

@drwestco There's no way to properly load the maps without encountering race conditions (I've resolved this in #192) and the map isn't initialized until AFTER the data is loaded, which is the primary blocker (the network latency/overhead, queuing to download `data.json`).

patricknelson:
I'll take that back; right now given how the code is structured and @r-pop's priority on this feature (as well as our current need to feature-freeze), it's best to avoid addressing that _right this moment_. Rebuilding this so that both can be kicked off asynchronously would take some time and the ethos of this site has been to be as rapid as humanly possible 😄 I've adopted that full force!

drwestco:
Does your latest change fix the problem in your screenshot, where "List of donation sites" appears? The getusppe.org frames scenario is map-only, since it's sized to push the list off the bottom of the frame. If the list header appears temporarily, it will look really strange.

patricknelson:
p.s. I realize I misread your statement (tired and going a bit too fast), I apologize @drwestco.

I will say it's easier right now to stick with the one loader instead of duplicating this 4x's, I'm "only" duplicating it 2x's and the logic is simpler too (since we would then also need to decide if we show the spinner only once or twice). If we show the spinner once per load, you need additional JS logic to only render inside the map placeholder.

> Does your latest change fix the problem in your screenshot, where "List of donation sites" appears? The getusppe.org frames scenario is map-only, since it's sized to push the list off the bottom of the frame. If the list header appears temporarily, it will look really strange.

Looks like it. See my override/screenshot below. Thanks for calling that out @drwestco, I wasn't aware that they were cropping the page content in the `<iframe>` 😄

patricknelson:
I will say I think it looks better centered...  Thoughts? Willing to do that only because it's a lot faster.

patricknelson:
Updated latest revision image in top comment.

patricknelson:
@r-pop rebased against `master` to include @alex's latest Heroku CI/CD updates.

public/locations-list-map.js
```diff
@@ -218,7 +216,10 @@ $(function() {
     const hideList = searchParams.get('hide-list') && searchParams.get('hide-list') === 'true';
 
     if (!hideList) {
-      $(".filters-list").html(createFiltersListHTML(data_by_location).join(" "));
+      // Generate and populate filter + list HTML. Once completed, swap loading and list container.
+      $('.filters-list').html(createFiltersListHTML(data_by_location).join(" "));
+      $('.locations-loading').hide();
+      $('.locations-container').show();
```
akainic: Do we want to show the map only if the list of states is shown, or regardless of if the list of states is shown? Just want to make sure we're not nesting the display of the map within this `if (!hideList)` check inadvertently!

patricknelson: I don't understand. The `hide-map` param already exists, so that's controlled independently (albeit not consistently, that too is also in #192 via https://github.com/r-pop/findthemasks/pull/192/files#diff-6e71b4786f0ea284c53245df40a9f6bdR252)

public/locations-list-map.js
akainic: Thanks for these thoughts! I just took a look at the new commit. My concern is that it doesn't quite solve the problem in the original issue, which is that we're not showing anything while we wait for the map to load. If we aren't making `$('.locations-container').show();` dependent on whether the map has actually loaded, do we have confidence that it has? I was just suggesting adding an event listener and conditionally doing `$('.locations-container').show();` based on the when the listener is fired. I think this would be fairly simple, but if it's too much to do in this PR I can try to fix this later!

public/locations-list-map.js
akainic: This code to hide the loading div and show the map div is nested under a check about whether to show the states list. I was wondering if that logically made sense. It seems like showing the map, as you described in mentioning the `hide-map` param, is independent from the `hide-list` param, and therefore shouldn't be nested within `!hideList`.

akainic: Oops! I didn't realize that the list container is separate from the map one. I see now that the loading work you added was just about the list and not about the map at all. Sorry for the confusion!

PR #221, opened 2020-03-26
Issue 197 : Map updates as filters change

drwestco:
Refactoring of data_by_location to store properties at the state+city level. Improved creation of DOM elements for list view. We now create these and the map markers up front, then just swap them as the filters change.

### DIFF
--- a/public/locations-list-map.js
+++ b/public/locations-list-map.js
@@ -1,8 +1,6 @@
 import toDataByLocation from './toDataByLocation.js';
 
 function createFiltersListHTML() {
-  $('.locations-container').show();
-
   const filters = [];
   filters.push(`<h4>${$.i18n('ftm-states')}</h4>`);
   for (const state of Object.keys(data_by_location).sort()) {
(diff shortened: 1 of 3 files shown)


public/locations-list-map.js
```diff
@@ -254,234 +263,183 @@ window.onFilterChange = function (elem, scrollNeeded) {
...
-  locationsListElement.innerHTML = htmlSnippets.join(" ");
+  const filters = { states, acceptItems };
+  const locationsList = $(".locations-list");
+
+  locationsList.empty().append(getFilteredContent(data_by_location, filters));
+  showMarkers(data_by_location, filters, false);
+
   if (scrollNeeded) {
-    locationsListElement.scrollIntoView({'behavior': 'smooth'});
+//    locationsList[0].scrollIntoView({'behavior': 'smooth'});
```
nburt: is this intentionally commented out?

krithin: If you intend to disable scrolling (which I agree is pretty jank right now) please call that out on the PR.

drwestco: Oops. I did that for testing, since it was seriously annoying while working on the filter updates. I'll put it back to preserve existing behavior, but do think this should be revisited.

drwestco: Restored the jank.

PR #223, opened 2020-03-26
Use CSS + semantic dom to show/hide the elements.

awong-dev:
Eliminates dependency on data.json XHR.
To productionize, there should be an index.php that takes the
data_fragment.html generated by the cloud function into the
location-list.


awong-dev:
https://findthemasks-semantic-d-sdi1zq.herokuapp.com/index-test.html

awong-dev:
This will also help with #208  because the donation list will show up immediately since it's already in the DOM.

drwestco:
This is a good idea, but my main concern here is maintainability. Consolidating index.html, give.html, and index.php+data_fragment.html, and on the script side: locations-list-map.js and locations-static.js. This should be done first before adding a new variant that will feature drift as independent devs continue to work on issues.

awong-dev:
Very good observation. If we were to merge this, my hope would be to delete index.html and locations-list-map.js once we've verified this all runs on the main site.

The question then just becomes merging give.html in.

patricknelson:
In plain terms, what is the primary goal (or set of goals) for this PR? I. Guessing speed up page load and centralize the two current pages, index + give, correct? Just confirming since it looks like the ticket starts out with implementation details (probably missing something).

awong-dev:
(1) load faster, esp on mobile.
(2) if we push a js bug (due to cache skew, etc) data still available.

patricknelson:
K, awesome, that’s a great idea. If I’m not mistaken, it may also really reduce the overhead of generating the HTML client side (with all the template interpolations and concatenation into the main list). Oh, plus you can take advantage of server-side `gzip` compression of HTML, so it still downloads quickly. 

Not at a computer to verify but: If not already, ensure you’ve got a bullet point to verify that the server’s configured (or CloudFlare) to gzip that content so it gets dumped down the pipe more quickly.

awong-dev:
cloudflare is already doing gzips and reencodes though I disabled the minify cause being able to read the code live was nice.

patricknelson:
Just choosing to log some of the results from my quick analysis here for posterity. Maybe not the best PR/issue to do this in, but I think this one has a lot of potential.

My conclusion is that out of the ~3.5s page render time (which includes the ~100ms-200ms network latency for `data.json`)*, about ~2.5s is spent inside `createContent` looping over 2k separate `addMarkerToMap` JS calls which involves lots of HTML parsing, then finally DOM render + CSS reflow.

So, I think the ideal solution appears to be a combination of multiple approaches for best performance:

1. Pre-render the HTML list here which will save lots of extra time currently delegated to the client right now in `createContent` (not including `addMarkerToMap`), see PR #223 (this ticket)
2. Delegate `addMarkerToMap` to end of `createContent` and try to add later all at once or ideally in chunks, see PR #274 for an approach to this

Also: Looking at `addMarkerToMap` there are some optimizations that could be performed there that may help with optimizing load speed, including:

- Calling `$.i18n` just once for each label ahead of time and caching labels for current language instead of calling it ~2,000 x 4  = 8,000 times
- Pre-linkify HTML in the `data.json` file (the [`linkifyHtml()` func](https://github.com/findthemasks/findthemasks/blob/master/public/locations-list-map.js#L715) called over and over for the marker pop-up)
- Create a single click listener function, named function in a higher scope, and then attach that to each marker (so less memory is used and potentially less CPU overhead when attaching listeners)

_\* Variables: I'm on a high speed cable connection (320mbit/s) with 4.3Ghz boost and it's still fairly CPU bound._

drwestco:
With the infinite scroll / partial loading, this is now obsolete.

patricknelson:
Oh my #396... that is a beautiful thing. It's so fast now! 🚀

PR #226, opened 2020-03-26
wip country param for i18n

nburt:
Still a WIP, I have not integrated with country-specific data.json files.

I have pseudo translated the site (?country=pseudo), see example: 
image

Todos:

1.  handle country-specific data.json files
2. Would like to understand how we're ingesting data from other countries a bit better. Generally the best practice would be to use locales rather than countries for i18n. For example we could use `es` for international spanish, `es-ES` for spain, `es-MX` for mexico as each country has slight differences in the language. For English `en-US` vs. `en-UK` have differences. It seems like we're trying to combine country specific data collection and map display with translation and they're slightly different. It is probably fine to just use "international/country" locales but would like to understand things a bit better. Related to this, our i81n file uses "en" in locales but if we're passing in countries maybe this locales constant should be updated to use country codes. Would also need to update links to different donation site forms based off the country
3. How to handle page navigation. If you're at /index.html?country=fr and navigate to /whoweare.html it should remain in French but currently reverts to english.


public/index.html
```diff
@@ -196,52 +196,51 @@
...
           <div id="locations-list" class="locations-list"></div>
         </div>
 
-        <p style="padding-top: 30px;">To add a donation site, update information, or request we take down information, please <a href="https://docs.google.com/forms/d/e/1FAIpQLSfgCpK5coPVFC6rJrE7ZhimiZuDoEaL6fo6gYqxsN_FIpJZhg/viewform">fill in this form</a>.<br><br>
+        <p style="padding-top: 30px;" class="i18n" data-i18n="[html]ftm-add-donation-site">To add a donation site, update information, or request we take down information, please <a href="https://docs.google.com/forms/d/e/1FAIpQLSfgCpK5coPVFC6rJrE7ZhimiZuDoEaL6fo6gYqxsN_FIpJZhg/viewform">fill in this form</a>.<br><br>
         Got a comment or suggestion not related to data updates? Want to help? contact@findthemasks.com or join us on <a href="https://github.com/r-pop/findthemasks">our Github</a>.</p>
-        <p class="centerText">made with &lt;3 in Seattle</p>
-        <a href="whoweare.html" class="end">who we are</a>
+      <p><span class='i18n' data-i18n="[html]ftm-made-with-love">Made with &lt;3 in Seattle by </span><a href="whoweare.html" class="end">findthemasks.com</a></p>
```
nburt: made a small change here for consistency between index/give

public/i18n.js
```diff
@@ -1,84 +1,179 @@
...
 
-$(function() {
-  $.i18n().load( locales );
+$(function () {
+  const searchParams = new URLSearchParams((new URL(window.location)).search);
+  const country = searchParams.get('country');
+
+  const defaultLocale = 'en';
+
+  $.i18n({ locale: (!!locales[country] && country) || defaultLocale }).load(locales);
```
raindrift: This looks like it will either use `defaultLocale` (en) or the `country` param. What we lose doing it this way is automatic browser locale detection. A better way might be something along the lines of:
```
const country = searchParams.get('country');
if(!!locales[country] && country) {
  $.i18n({ locale: country }).load(locales);
} else {
  $.i18n().load(locales);
}
```

It would also work to fetch the browser language here instead of letting the library do it, but I don't think that's quite as good: https://stackoverflow.com/questions/673905/best-way-to-determine-users-locale-within-browser

raindrift:
> For example we could use es for international spanish, es-ES for spain, es-MX for mexico as each country has slight differences in the language. For English en-US vs. en-UK have differences.

jQuery.i18n is able to handle this gracefully. I believe if we create `en-gb`, for example, that it will be used when that is the browser's locale setting, but if the user has `en-ca` set and we don't support that, it'll default to `en`.

nburt:
Excuse the massive brain dump, would love input from others here. Just putting down all the solutions I can think of.

**My understand of the i18n problem is this:**

- Data ingestion. We want local / country-specific moderators to manage data moderation. These teams should be able to operate in isolation.
- We want users from around the world to be able to access findthemasks.com in their native tongue.
- We want users from around the world to see the map filtered to the closest marker to them. If they access the site from France, they should see French donation sites by default.

**Brain dumping a few different solutions, would love for you all to chime in on pros/cons**

- **Separate forms for each country (current plan)**
  - This allows for isolation in data moderation
  - This makes displaying markers as you go between countries challenging (is this a use case?)
  - Poses challenges for i18n. The current plan is to use a country query param for both locale selection + data selection. What if I’m French but live in the US? If I live in Canada (country=ca) how do we decide which locale to load?
- **One master form**
  - This simplifies the marker problem
  - Could cause performance issues
  - Will cause data moderation challenges because you’d be filtering through thousands of rows
  - Allows us to separate the locale problem from how we load data
- **Separate forms that are somehow merged during deploys**
  - Moderation can happen separately
  - More complexity during deploys
  - All forms would need to be consistent
  - Allows us to display markers as users scroll around the world.
  - Could cause performance issues if we’re loading a really large data.json file, would need to make updates to filter the data by user’s country for map markers + location lists (but we pretty much need to do this anyways)
  - Allows us to separate the locale problem from how we load data
- **Switch to a server framework / database**
  - Benefits of separate forms merged during deploys (single data source) but with better performance
  - Adding automated tests probably is a bit easier with frameworks
  - Data moderation process would change but could happen country by country. I’m most familiar with Ruby/Rails but I don’t think it’d be hard to CRUD “donation sites” or use a library like RailsAdmin for this. May need to add some basic user permissions or an admin section (hardcoded username/password?) so data moderation isn't publicly available)
  - Time to set up / downtime when switching over?
  - Additional complexity, probably makes things a little less accessible
  - Server infrastructure, etc.
  - Could stand up APIs to make the data more accessible to other orgs

public/i18n.js
nburt: yes could call, I tested and that works much better. fr-FR as a param would falls back to fr. Garbage as the param falls back to english

raindrift:
Here are my thoughts on this:

I think the right thing here is to keep browser language and location
selection entirely separate. This is what other apps do, and it’s correct.
The language someone speaks and the place where they’re standing are only
loosely correlated. :)

Language selection should pick the translation based on browser locale,
with a language selector for people whose browsers are messed up.

The map should be centered on the place where the user is located. If we
don’t get location data from the browser, we could try geoip or we could
just ask them to pick a country.

Managing the data is a separate issue from the UX. I think that for now,
merging all the data from the different country forms into a single
data.json is the most expedient, and then we can move to a more granular
solution for loading that data separately.

Dispatching the user to the right form should happen based on the location
they gave us. Sometimes people will end up at the wrong one, and then
volunteers will have to send them to the right one. The other option is for
us to have a single form and divide the data by country on the backend as
it comes in. I’m not sure which of these is better at this point.

graemer:
Agree on language and location being separate. Use the browser default for lang and we're likely good enough.

Separating LOCATION is the critical thing at this point. We should assume that health systems / needs / moderators will be almost entirely divided on national boundaries.

I think that we need to always have a "context" for the country, which defines which data is available / applicable, and which form the user is directed to.

My suggestion would be to be explicit, and include this in the URL as a param (and we could build a bit of UI to switch between countries).

As far as merging the data.json's I have significant perf worries in the short term, and getting something done and available TODAY is critical.

We're already stressing page perf if we're loading just the US data, and if we start adding thousands of other points that are largely irrelevant, we just hurt individual users.

I do like the idea of standing up a country=all option so that we can demo really well, but I think for actual usage we should restrict to just a country at a time.

public/index.html
```diff
@@ -191,57 +193,82 @@
...
+          <a class="dropdown-item" href="?locale=en">English</a>
+          <a class="dropdown-item" href="?locale=fr-FR">French (France)</a>
+        </div>
+      </li>
+      <li class="nav-item dropup">
+        <a class="nav-link dropdown-toggle" href="#" id="countries-dropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+          Country
+        </a>
+        <div class="dropdown-menu dropdown-menu-right" aria-labelledby="countries-dropdown">
+          <a class="dropdown-item" href="/fr/index.html">France</a>
```
nburt: need to figure out how we maintain the locale when switching between countries / pages in general. Should we use local storage to write the user preference?

public/fr/.htaccess
```diff
@@ -0,0 +1,25 @@
+<ifModule mod_rewrite.c>
+    RewriteEngine On
+    RewriteCond %{REQUEST_FILENAME} !-f
+    RewriteCond %{REQUEST_FILENAME} !-d
+    RewriteRule (.*) index.html [QSA,L]
```
nburt: Need to verify that this will work

public/i18n.js
```diff
@@ -1,84 +1,181 @@
...
+  }
+};
 
-$(function() {
-  $.i18n().load( locales );
+$(function () {
+  const searchParams = new URLSearchParams((new URL(window.location)).search);
+  const locale = searchParams.get('locale');
+
+  if (locale) {
```
nburt: try locale, if not default to user browser default

nburt:
Closing and will move to new pr as project member to get heroku previews

PR #242, opened 2020-03-26
Favicon and meta in <head>

cjimmy:
* Moved favicons to subfolder, adjusted *.html accordingly.
* added <meta> og tags for unfurl preview
* added canonical url in head


graemer:
@cjimmy - Is it not possible to move the french-specific strings within the meta tags to https://github.com/findthemasks/findthemasks/blob/master/public/i18n.js? (you'd likely need to extend a bit)

cjimmy:
Good question. After looking into it, I suspect we can't if we want crawlers to parse it correctly. Most crawlers take the html as it is without rendering the js. Google does render js when it crawls, but I don't know if jquery.i18n will correctly localize the head tags in time. Facebook's crawler doesn't even render js even though they own React and are responsible for single-page apps being popular. It's something we should test, especially if we plan to internationalize more. For right now, my rec is to hardcode it for SEO sake, and test how if it works in i18n.js.

graemer:
Yeah, I suspect we'll need to move to server-side rendered if we want to do this...

awong-dev:
Don't forget we have a php backend

PR #244, opened 2020-03-26
Dedup cities in parsed data

drwestco:
Raw data doesn't have consistent case in the Cities field, so we can wind up with duplicate entries in the rendered list.


public/toDataByLocation.js
```diff
@@ -3,14 +3,13 @@ export default function toDataByLocation(data) {
   const approvedIndex = headers.findIndex( e => e === 'approved' );
   const stateIndex = headers.findIndex( e => e === 'state' );
   const cityIndex = headers.findIndex( e => e === 'city' );
-  const instructionsIndex = headers.findIndex( e => e === 'instructions' );
```
graemer: Why is this removed?

drwestco: Cleanup from previous checkin. Linkify code used to be done here; now this value is unused.

PR #253, opened 2020-03-27
Use ❤️ instead of <3

patricknelson:
Just thought ❤️ looked nicer 😄 

----

**vs**

image


r-pop:
I went with text over emoji on purpose inititally, but I defer to @cjimmy's design judgement.

patricknelson:
Fair enough, just throwing it out there (I don't have a strong opinion either way). 😃

cjimmy:
I had the same thought as you Patrick, but if @r-pop did it on purpose, I want to respect that. I think the <3 does a good job showing the scrappiness of the team.

PR #258, opened 2020-03-27
Dynamic maps key: Proof of concept method of loading google maps with a dynamic API key.

patricknelson:
Just a proof of concept method of loading google maps with a dynamic API key. This is very hacky right now due to the inter-dependencies of the rest of the code that rely on maps even though technically the map may actually be hidden. So, right now we **must** load the maps API even if the map isn't displaying, so this has to cope with that by loading maps at the root module context.

Use `?mapsApi=[your key]` to override the maps key.

Since this is a hack, I don't necessarily expect this to be merged but wanted to put it out there as inspiration for a better executed version! 😄 It will become increasingly necessary as we deploy more map updates to Heroku (which is not capable of loading our maps API with special wildcards like `findthemasks-*.herokuapp.com/*` given limitations on Google's side). 

Clone `git@github.com:patricknelson/findthemasks.git` and checkout `dynamic-maps-api` to continue with this code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ Join the slack! <https://join.slack.com/t/findthemasks/shared_invite/zt-czdjjznp
 ?hide-map={true/false}
 ?hide-filters={true/false}
 ?hide-list={true/false} (also hides filters)
+?mapsKey={custom Google Maps API key}
 ```
 
 All boolean parameters default to false.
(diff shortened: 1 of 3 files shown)


patricknelson:
Going to delete this pile of trash 😄 

🔥🗑

PR #263, opened 2020-03-27
Map search: Fix for no results on enter (without pressing down).

patricknelson:
Ensuring we attempt to geocode the raw user entered text if they don't press down arrow to select an entry from Google's autosuggest. The solution is to just directly geocode the user input text, cross fingers 🤞 and go with Google's top hit.

**Here's how it looks now:**

### DIFF
--- a/public/locations-list-map.js
+++ b/public/locations-list-map.js
@@ -417,18 +417,49 @@ function initMapSearch() {
     return;
   }
 
+  // Search element (jquery + html element for autocompleter)
+  const $search = $('#map-search'),
+        searchEl = $search[0];
+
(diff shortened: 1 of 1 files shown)


public/locations-list-map.js
```diff
@@ -460,32 +491,6 @@ function resetMap() {
 
 
 
-/**
- * Takes the form input and attempts to pull geocoded information from it. Once done, centers map on closest markers nearby.
- */
-function loadSearchInMap() {
```
patricknelson: Removed this function because I needed the scope from the caller so I could pull the `$search.val()` without duplicating too much code.

PR #265, opened 2020-03-27
Clean up cloud funcs. Support fr and ch.

awong-dev:
With this, the pipeline now generates a data-{country}.json
file. A symlink from data.json -> data-us.json is provided
for compatibility.

It also includes some cleanups of the data.json such that
unlabeled rows do not appear.


awong-dev:
TBR=@csakoda , @graemer

Issue #271, opened 2020-03-27
Make Geolocation call async 


bamnet:
Currently when the map is loading it calls out to the HTML5 Geolocation API to detect the user's location and center the map appropriately.  This is in the blocking path of displaying all the pins. As a result, users need to interact and Approve/Block/Dismiss the browser notification before the map finishes loading.

New users aren't particularly likely to interact with this prompt and the map just stalls out grey until they do.

The map should have a center and zoom set before a browser geolocation call is made and, ideally, that call should happen after all the heavy processing of the pins, etc has occured.

drwestco:
Search bar + "use my location" replaces this? I think the old geolocation on init code can be completely removed now.

PR #272, opened 2020-03-27
temp disable user geolocation

nburt:


nburt:
Just a disable for now, followup tracked here: https://github.com/findthemasks/findthemasks/issues/271

PR #273, opened 2020-03-27
Center the map on the US by default.

bamnet:
This is a very quick workaround fo #271.

It doesn't actually make the geolocation call non-blocking, but gives the map an initial center and zoom so it will render markers and show a map instead of a grey box while the geolocation prompt is still open.


bamnet:
I'll walk this back in favor of @nburt's quicker fix in #272.  I can spend some cycles this afternoon (eta 2-3 hours from now) cleaning up the geolocation call.

Issue #270, opened 2020-03-27
Add marker clustering and progressively update markers on zoom


nburt:
Currently we render all markers at once. This creates a pretty much unusable default US view because we are rendering 1800+ markers. We should integrate marker clustering for a better user experience: https://developers.google.com/maps/documentation/javascript/marker-clustering

dhruv4:
I can hop on this

PR #274, opened 2020-03-27
Marker cluster

dhruv4:
Added marker clustering on map load. It should make it a bit faster and a bit more digestible.

Referencing issue #270


public/locations-list-map.js
```diff
@@ -618,14 +618,16 @@ function showMarkers(data, filters, showNearest) {
       }
     }
   }
+  var markerCluster = new MarkerClusterer(map, markers,
+    { imagePath: 'images/markercluster/m' });
```
drwestco: How is the markerCluster instance managed? showMarkers is called multiple times as the filters change. Need to tear down the previous cluster and create a new one?

drwestco: I think you can create it once when the map is created, then just call clearMarkers(), followed by addMarkers() as the filters change.

public/locations-list-map.js
```diff
@@ -609,7 +609,7 @@ function showMarkers(data, filters, showNearest) {
         if (marker) {
           if (inStateFilter && inAcceptFilter) {
             markers.push(marker);
-            marker.setMap(map);
+            // marker.setMap(map);
```
drwestco: Remove these lines - don't comment them out.

public/locations-list-map.js
```diff
@@ -708,32 +710,32 @@ function centerMapToNearestMarkers(map, markers, fallbackBounds) {
...
+    `<div class="label">${$.i18n('ftm-maps-marker-address-label')}</div><div class=value>` + address + '</div>' +
+    `<div class="label">${$.i18n('ftm-maps-marker-instructions-label')}</div><div class=value>` + linkifyHtml(instructions) + '</div>' +
+    `<div class="label">${$.i18n('ftm-maps-marker-accepting-label')}</div><div class=value>` + accepting + '</div>' +
+    `<div class="label">${$.i18n('ftm-maps-marker-open-packages-label')}</div><div class=value>` + open_accepted + '</div>';
+
+  var location = { lat: latitude, lng: longitude };
+  var marker = new google.maps.Marker({
+    position: location,
+    title: name,
+    // map: map
```
drwestco: Remove this line too. And can remove the 'map' param. Maybe rename method to 'createMarker', since the add operation has been moved to showMarkers since a day or so ago.

dhruv4:
@drwestco thanks for the comments!! I've updated the branch with those changes.

public/locations-list-map.js
```diff
@@ -570,11 +572,15 @@ function centerMapToMarkersNearCoords(latitude, longitude) {
  * Changes the markers currently rendered on the map based strictly on . This will reset the 'markers' module variable as well.
  */
 function showMarkers(data, filters, showNearest) {
+
   markers = [];
 
   if (!map) {
     return;
   }
+  if (markerCluster){
```
drwestco: If you're worried about markerCluster not existing, add it to the 'if (!map)' check up above. You're using markerCluster blindly down below when adding the markers, so checking it like this doesn't buy you anything.

graemer: @drwestco suggest a change?

dhruv4: fixing this rn!

PR #285, opened 2020-03-27
Fix cors for data files

awong-dev:


awong-dev:
@alex This is hack version. It'd be nice to have a pattern. Do you. know how that might work?

alex:
@awong-dev https://httpd.apache.org/docs/2.4/mod/core.html#filesmatch

PR #302, opened 2020-03-28
Move translations to json so smartling can expand them easily

raindrift:
Stores translations as JSON files and loads them via xhr. This is a better way to do it, but we didn't before because we weren't running a dev webserver, and xhr doesn't work with local files.

Since translations are now retrieved later in the page rendering pipeline, it is necessary to block display of the page until they're loaded. Otherwise the page renders in English (the fallback in the dom) and then switches languages a moment later, which is very disorienting.

So, now the page shows a spinner until translations are loaded. Loading `data.json` happens in parallel, so the time to a usable page is still about the same.

Also switched to bootstrap spinners because they look better when rendered in the center of the page.


raindrift:
Note: the following French translations have been updated using google translate, since they were missing:
```
    ftm-add-donation-site-answer
    ftm-add-donation-site-question
    ftm-large-donation-answer
    ftm-large-donation-question
```

raindrift:
This should be ready to go now.

public/give.html
```diff
@@ -305,54 +305,71 @@
...
 
 </head>
 
 <body>
 <main class="content">
-    <div class="container" style="padding-bottom: 40px;">
-      <div class="map-container" style="display: none;">
+    <div class="languages-loading" style="position: fixed;top: 50%; left: 50%;">
+        <div class="spinner-border" role="status">
+          <span class="sr-only">Loading translations...</span>
```
nburt: should this be translated as well?

public/index.html
```diff
@@ -570,146 +574,156 @@
...
-        </ul>
-      </div>
-    </nav>
-      <div class="heading-container">
-        <img class="logo" src="images/logo-color.svg" alt="Find The Masks logo" />
-        <!-- <h1>#findthemasks</h1> -->
-        <div class="member-of i18n" data-i18n="[html]ftm-index-site-coalition">A member of the <a target="_blank" rel="noreferrer noopener" href="https://getusppe.org/">GetUsPPE.org</a> coalition</div>
+    <div class="languages-loading" style="position: fixed;top: 50%; left: 50%;">
+        <div class="spinner-border" role="status">
+          <span class="sr-only">Loading translations...</span>
```
nburt: same here

public/index.html
```diff
@@ -570,146 +574,156 @@
...
-          <div class="i18n" data-i18n="ftm-add-donation-site-question"> Want to add a donation site? update or remove information?</div>
-          <div class="i18n" data-i18n="[html]ftm-add-donation-site-answer">
-            Please <a href="https://docs.google.com/forms/d/e/1FAIpQLSfgCpK5coPVFC6rJrE7ZhimiZuDoEaL6fo6gYqxsN_FIpJZhg/viewform">
-            fill out this form</a>. We will review it as soon as possible.
+        <div class="content-container" style="margin-bottom: 40px;">
+          <h2 class="i18n" data-i18n="ftm-protecting-healthcare-workers">Protecting healthcare workers protects everybody</h2>
+
+          <p class="i18n" data-i18n="ftm-index-where-to-find-ppe">
+            Maybe you bought N95 masks a few weeks ago or already have some in your house. Right now healthcare workers are running dangerously low on PPE like masks, hospital gowns, booties, and face shields to protect them while they risk their lives to save others.
+          <p>
```
nburt: I think this was already there but open/close tags look a little off

public/index.html
```diff
@@ -570,146 +574,156 @@
...
+          </ol>
 
-        <div class="question i18n" data-i18n="[html]ftm-large-donation-question">Have <em>a lot</em> (500+ items) to donate?</div>
-        <p class="i18n" data-i18n="[html]ftm-large-donation-answer">We encourage large donors to use another site we work with to help route donations. Register at <a target="_blank" rel="noreferrer noopener" href="https://getusppe.org/give">getusppe.org/give</a>.</p>
+          <div class="question i18n" data-i18n="[html]ftm-large-donation-question">Have <em>a lot</em> (500+ items) to donate?</div>
+          <p class="i18n" data-i18n="[html]ftm-large-donation-answer">We encourage large donors to use another site we work with to help route donations. Register at <a target="_blank" rel="noreferrer noopener" href="https://getusppe.org/give">getusppe.org/give</a>.</p>
 
-        <div class="question">Can't find a donation site near you?</div>
-        <p>Visit <a target="_blank" rel="noreferrer noopener" href="https://getusppe.org/give">getusppe.org</a> to let them know what you have available.</p>
+          <div class="question">Can't find a donation site near you?</div>
```
nburt: can you add i18n here?

image

public/index.html
```diff
@@ -570,146 +574,156 @@
...
+            <div class="map-search-wrap">
+              <div class="search-input">
+                <input type="search" name="map-search" id="map-search" placeholder="Search by city, zip, or address" data-i18n="[placeholder]ftm-maps-search-placeholder">
+                <!-- Icon credit to https://feathericons.com/  -->
+                <img class="search-icon" src="images/icons/search.svg">
+              </div>
+              <!-- BETA: To enable search field, either change data-enable-search to "true" in #map below, or add hide-search=false/show-search=true to query string. -->
+              <div class="search-controls">
+                <a href="javascript:;" class="map-action-link" id="use-location"><span data-i18n="ftm-maps-search-use-location"></span>
+                  Use my location<span class="search-control-separator">|</span></a>
```
nburt: Span is a little off here

image

public/give.html
raindrift: Not this one. It appears before translations are loaded, so it can't be. I am hoping that the number of non-English-speaking blind users we have will be small enough that it'll be ok. :)

public/index.html
raindrift: Same

public/index.html
raindrift: wow wtf :)

public/index.html
raindrift: Yep!

public/index.html
raindrift: yep, looks like it

PR #305, opened 2020-03-28
add bottom nav for ca/fr/us

nburt:
Language + country selector for whenever we are ready to turn on France / Canada

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -315,7 +315,33 @@
 </head>
 
 <body>
+<nav class="navbar fixed-bottom navbar-expand-lg navbar-light bg-light">
+  <a class="navbar-brand" href="#">#findthemasks</a>
 
+  <div class="ml-auto">
+    <ul class="navbar-nav">
+      <li class="nav-item dropup">
+        <a class="nav-link dropdown-toggle" href="#" id="locales-dropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+          Language
+        </a>
+        <div class="dropdown-menu dropdown-menu-right" aria-labelledby="locales-dropdown">
+          <a class="dropdown-item" href="?locale=en">English</a>
+          <a class="dropdown-item" href="?locale=fr-FR">French (France)</a>
+        </div>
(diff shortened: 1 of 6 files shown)


drwestco:
A little strange that picking "France" from the country selector leaves me on a page still in English.  Also, need some indicator of current language+country?

public/index.html
```diff
@@ -315,7 +315,33 @@
 </head>
 
 <body>
-
+<nav class="navbar fixed-bottom navbar-expand-lg navbar-light bg-light">
+  <a class="navbar-brand" href="#">#findthemasks</a>
```
drwestco: Why is this a link? Placeholder that's eventually going to navigate somewhere?

nburt:
@drwestco yeah sorry, not ready yet I shouldn't opened it. Realized what you commented and working on a js solution to ensure that country + locales stay the same as you navigate

public/index.html
```diff
@@ -315,7 +315,33 @@
...
+
+  <div class="ml-auto">
+    <ul class="navbar-nav">
+      <li class="nav-item dropup">
+        <a class="nav-link dropdown-toggle" href="#" id="locales-dropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+          Language
```
drwestco: "Language"/"Country" labels should be translated.

drwestco: As should the popup entries.

nburt:
there is a flash of English when in a foreign language and switching between pages still, that will be fixed with this: https://github.com/findthemasks/findthemasks/pull/302

public/locations-list-map.js
```diff
@@ -35,6 +36,39 @@ function GetCountry() {
   return 'us';
 }
 
+const generateBottomNav = () => {
+  const localeDropdown = document.getElementById('locales-dropdown-selector');
+  const countryDropdown = document.getElementById('countries-dropdown-selector');
+
+  if (localeDropdown && countryDropdown) {
```
nburt: only render on the index.html page

PR #308, opened 2020-03-28
Add filters state object

### LINKED ISSUE #304: State filter is broken for France
image

drwestco:
Revamp of how the active filters are tracked, for both map and list. There's now a data object that tracks the available filters and their current states. The UX elements are created based on this object (or not, if hide-list and/or hide-filters are set). State is synced between the UX elements and any ?state= params on page load.

Fixes some jank in onFilterChanged, especially WRT calling it directly at startup to handle ?state= settings.

Also simplifies the "Reset map" behavior, since it can just revert to what the filters object holds, instead of tracking the initial filters separately.

Fixes #304

### DIFF
--- a/public/locations-list-map.js
+++ b/public/locations-list-map.js
@@ -15,8 +15,6 @@ let markerCluster = null;
 // Configuration defined in query string. Initialized in jQuery DOM ready function.
 let showMapSearch = false; // BETA FEATURE: Default to false.
 
-let initial_marker_filters = null; // NOTE: Defined in map initialization function.
-
 // Keep track of the previous info windows user has clicked so we can close them.
 let openInfoWindows = [];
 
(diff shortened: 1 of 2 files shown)


public/locations-list-map.js
```diff
@@ -276,94 +304,69 @@ $(function () {
 
     showList && createContent(window.data_by_location);
 
-    const stateFilter = {};
-    let hasStateFilter = false;
+    const filters = createFilters();
+
+    // Update filters to match any ?state= params
+    const stateParams = searchParams.getAll('state').map(state => state.toUpperCase());
+    const states = stateParams.map(param => param.split(',')).reduce((acc, val) => acc.concat(val), []);
```
awong-dev: nit: I wonder if this would be faster with a spread operator and push rather than concat?

public/locations-list-map.js
```diff
@@ -389,49 +392,36 @@ function loadMapScript(searchParams, stateFilter) {
...
-  var data_by_location = window.data_by_location;
-
-  var element = document.getElementById('map');
+function initMap(filters) {
+  const element = document.getElementById('map');
 
-  if (element == null) {
-    alert('could not find map div');
+  if (!element ) {
+    return
```
awong-dev: nit: semicolon and extra space after element.

awong-dev: fixed.

PR #312, opened 2020-03-28
Redesigned frontend

cjimmy:
## Index.html redesign
Screen Shot 2020-03-28 at 1 09 58 AM
Screen Shot 2020-03-28 at 1 11 07 AM

**Please, please check this out before merging and check that the maps still works.** It's something that wasn't working for me in dev, and I can't verify that it works.

## Included in this change
* Complete redesign + copy rewrite of front page
* associated changes to `en` of i18n, but not `fr` or `ca` (It will simply revert to prev text, which is fine)
* new images

### Not included
* Code does not touch any other page (e.g.`give.html`, `stats.html`, or `whoweare.html`)
* changes to social unfurl (like image)


alex:
Very small note: The count below "Donation Sites" (the "(X so far!)"), this reflects the count after filters have been applied, which I think can be a bit confusing. I don't have a particular suggestion -- I just noticed this after applying a filter, then scrolling back to the top of the page, and was confused :-)

cjimmy:
Didn't know that! Will remove and revert it back to just a number.  Thanks @alex

cjimmy:
Holy crap how did you get the handle @alex

nburt:
Noticing something with i18n for the states

US:
image

France (no abbreviations)
image

Is this something that is just due to France data not being correct? Not sure if we can assume 2-letter short codes with each state, départment, etc. around the world

nburt:
Did you leave give.html as is since it is embedded in other sites?

cjimmy:
Good catch @nburt. It's an issue because I assumed it would be two letters – I didn't think about France. Not an issue with data. I'll have to move it to the top. **@ anyone**,  what's the best way in index.html to know what language/country is selected.

cjimmy:
> Did you leave give.html as is since it is embedded in other sites?

@nburt Just not the priority at the moment, but that will be next. Didn't want to be changing css in two places if we have yet to nail down the first.

nburt:
@cjimmy it's in the url: findthemasks.com/fr/index.html, findthemasks.com/us/index.html, findthemasks.com/ca/index.html, etc.

cjimmy:
actually I think the best solution is to just change the font size: 
Screen Shot 2020-03-28 at 11 43 26 AM

nburt:
> actually I think the best solution is to just change the font size:
> Screen Shot 2020-03-28 at 11 43 26 AM

This will work to get things merged, we'll need to keep an eye on it though. If the states get any longer won't it overflow?

public/i18n.js
```diff
@@ -102,19 +102,21 @@ const locales = {
...
+    "ftm-index-types-of-ppe": "If you know construction workers, vets, scientists, or dentists, please share this link with them.",
+    "ftm-what-to-do-header": "How to Donate PPE",
+    "ftm-what-to-do-1": "<b>Use our map</b> below to find a hospital or health center near you accepting donations.",
+    "ftm-what-to-do-2": "<b>Follow their instructions</b> for what they accept and how to drop it off.",
+    "ftm-what-to-do-3": "<b>If you are donating opened packages</b>, please put them in a ziploc bag with a note saying whether they have been used, whether anyone in your household has been sick or travelled outside the country, and the expiration date if known.",
+    "ftm-add-donation-site-question": "Want to add a donation site? update or remove information?",
+    "ftm-add-donation-site-answer": "Please <a href='https://docs.google.com/forms/d/e/1FAIpQLSfgCpK5coPVFC6rJrE7ZhimiZuDoEaL6fo6gYqxsN_FIpJZhg/viewform'>fill out this form</a>. We will review it as soon as possible.",
```
nburt: the donation form in the string needs to be `<a class=\"add-donation-site-form\">fill in this form</a>` to redirect correctly to the correct country form using this js: https://github.com/findthemasks/findthemasks/blob/39998ef9cbf3daacd07a5a9d3b42d00abcf7a1b1/public/locations-list-map.js#L307

cjimmy: thanks

cjimmy: so it should be 
```
"ftm-add-donation-site-answer": "Please <a class=\"add-donation-site-form\">fill in this form</a>. We will review it as soon as possible.",
``` ???

nburt: 👍 yup, exactly

cjimmy:
> This will work to get things merged, we'll need to keep an eye on it though. If the states get any longer won't it overflow?

How about breaking the word? not ideal but will work for now until we know how long these names will be
Screen Shot 2020-03-28 at 11 52 52 AM

nburt:
> > This will work to get things merged, we'll need to keep an eye on it though. If the states get any longer won't it overflow?
> 
> How about breaking the word? not ideal but will work for now until we know how long these names will be
> Screen Shot 2020-03-28 at 11 52 52 AM

This is the list for France: https://en.wikipedia.org/wiki/List_of_French_departments_by_population, can you edit the html to check out a few of the longer ones and see what it looks like?

public/i18n.js
cjimmy: change committed

patricknelson:
Just wanted to say: This looks really good! Great work!

Issue #309, opened 2020-03-28
Keep original line breaks in "Instructions" field


cchang-adi:
The original submissions for many listings include line breaks in the "Instructions" field. For example, row 1140 (aka our first Alaska submission!): 
![image](https://user-images.githubusercontent.com/45407457/77817901-96251880-708b-11ea-894b-5ad21d434eb0.png)
(top paragraph is cut off since we don't have word wrap in cells)

However, the existing page does not honor line breaks, so the result on our webpage is this: 
![image](https://user-images.githubusercontent.com/45407457/77817915-ab9a4280-708b-11ea-8a80-a8f704217f11.png)

The latter is unreadable and cluttered. 

Is there a way to honor line breaks in the "Instructions" field as they appear in the form responses?

PR #324, opened 2020-03-28
Preserve original linebreaks in the instructions strings.

krithin:
Resolves https://github.com/findthemasks/findthemasks/issues/309

### DIFF
--- a/public/locations-list-map.js
+++ b/public/locations-list-map.js
@@ -182,17 +182,34 @@ function createFilterElements(filters) {
 
 // Wrapper for document.createElement - creates an element of type elementName
 // if className is passed, assigns class attribute
-// if child is passed, appends to created element
+// if child (either a node or an array of nodes) is passed, appends to created element.
 function ce(elementName, className, child) {
   const el = document.createElement(elementName);
   className && (el.className = className);
-  child && el.appendChild(child);
+  if (child) {
+    if (Array.isArray(child)) {
+      child.forEach((c) => el.appendChild(c));
+    } else {
+      el.appendChild(child);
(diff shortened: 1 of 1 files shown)


public/locations-list-map.js
```diff
@@ -195,6 +201,17 @@ function ctn(text) {
   return document.createTextNode(text);
 }
 
+// Turns a string with embedded \n characters into an Array of text nodes separated by <br>
+function multilineStringToNodes(input) {
+  const textNodes = input.split('\n').map((s) => ctn(s));
```
drwestco: Stupid jscript trick. You can eliminate the foreach down below and turn this into a one-liner:
```
return [].concat(...input.split('\n').map((s) => [ ctn(s), ce('br') ]));

File: public/locations-list-map.js [hunk 2]
```diff
@@ -195,6 +201,17 @@ function ctn(text) {
...
+  let returnedNodes = [];
+  textNodes.forEach((e) => {
+    returnedNodes.push(e);
+    returnedNodes.push(document.createElement('br'));
+  })
+  return returnedNodes.slice(0,-1);
```
drwestco: You can just decrement the length of returnedNodes to chop off the last <br>, instead of making a copy via slice()

krithin:
I'm afraid I wasn't able to get to this yesterday and won't today either,
so feel free to push commits to this branch if you'd like! And if not I
will be able to do something here tomorrow.

awong-dev:
@drwestco  if you'd like to make the efficiency fixes, please do so in a follow-up. Merging this.

Issue #317, opened 2020-03-28
Start tracking user actions in google analytics


saikatbhadra:
## Developer Story
As a developer I want to see which actions my users take.

## Acceptance Criteria (WIP)
An action in GA in generated when:

### Map actions

- [x] A user initiates a map search
- [x] A user clicks a state filter
- [x] A user clicks an equipment filter
- [x] A user resets the map
- [x] A user clicks on "use my location"

### i18n
- [x] A user changes country
- [x] A user changes language

saikatbhadra:
Taking this issue on

saikatbhadra:
Closing this issue as the major events are covered. If anyone wants more GA events please open a new issue

PR #329, opened 2020-03-28
Add GA tracking for clicking map filters

saikatbhadra:
Adds tracking for the equipment and state map filters.

Everytime a map filter is created created an analytics event with these properies:
**Category**: `map`
**Action**: `filter-getItems` or `filter-states`
**Label**: key of the filter clicked (i.e. `WA` or `mask`)


saikatbhadra:
PR related to: https://github.com/findthemasks/findthemasks/issues/317

PR #333, opened 2020-03-29
Fix missing strings, add Italian and German to locales dropdown

nburt:


public/i18n/it-it.json
```diff
@@ -24,6 +24,8 @@
   "ftm-filters" : "Filtri",
   "ftm-france" : "Francia",
   "ftm-french" : "francese",
+  "ftm-italian" : "italiano",
+  "ftm-german" : "Tedesco",
```
krithin: Tiny nit: inconsistent capitalization across language names.

nburt: this is google MT, I'll update

nburt: Going to merge as is but I edited it in Smartling so it'll come through soon

PR #344, opened 2020-03-29
Minor css fixes

cjimmy:
* makes index.html a little more mobile-responsive. Makes images smaller, per @saikatbhadra 's concerns.
* centers "Loading Donation Sites..." with spinning loader
* addresses issue @nburt found with longer state names. i.e. see 
Screen Shot 2020-03-29 at 3 59 03 PM
(This is to show that in `en` it'll look as is, and in `fr`, it'll push the content to the next row. We probably won't see situations where we have both 2-char **and** greater-than-2-char on the same page.)

---------
* moves map stats number (number of donation sites) inline with Donation Locations header as before, but a less prominent.
Screen Shot 2020-03-29 at 4 01 19 PM


public/index.html
```diff
@@ -649,19 +708,15 @@ <h2 class="i18n" data-i18n="ftm-protecting-healthcare-workers">Protecting health
...
-
-          <h2 class="i18n" data-i18n="ftm-add-donation-site-header">Do you need PPE?</h2>
           <div class="need-ppe">
-            <div class="i18n" data-i18n="ftm-add-donation-site-question"> Want to add a donation site? update or remove information?</div>
+            <h2 class="i18n" data-i18n="ftm-add-donation-site-header">Do you need PPE?</h2>
+            <div class="i18n" data-i18n="ftm-add-donation-site-question"><p>Want to add a donation site? update or remove information?</p></div>
             <div class="i18n" data-i18n="[html]ftm-add-donation-site-answer">
-              Please <a href="https://docs.google.com/forms/d/e/1FAIpQLSfgCpK5coPVFC6rJrE7ZhimiZuDoEaL6fo6gYqxsN_FIpJZhg/viewform">
-              fill out this form</a>. We will review it as soon as possible.
+              <p>Please <a href="https://docs.google.com/forms/d/e/1FAIpQLSfgCpK5coPVFC6rJrE7ZhimiZuDoEaL6fo6gYqxsN_FIpJZhg/viewform">
```
nburt: since this is an html string you also need to update it in `en.json` to add the <p> tags. The content here is just fallback in case the correct language can't be loaded. As is this will render a <p> tag for fallback but not for other languages.

cjimmy: Right. Thanks.

cjimmy:
Add this commit: 
Screen Shot 2020-03-29 at 4 14 29 PM

graemer:
@nburt - could you take a shot at resolving the es & pl JSON issues?

graemer:
They look simple enough, but I'm not completely on top of what the implications are when making changes to these translation files.

PR #362, opened 2020-03-30
update canada national donation site cta

nburt:


nburt:
hold off merge until translations come through

PR #371, opened 2020-03-31
Only fetch the locale file we need on page load

nburt:


public/i18n.js
```diff
@@ -11,10 +11,45 @@ const locales = {
...
+
+const determineLocaleConfig = (detectedLocale) => {
   const searchParams = new URLSearchParams((new URL(window.location)).search);
-  const locale = searchParams.get('locale');
+  const localeParam = searchParams.get('locale');
+
+  // first try locale selected in app by user
+  // next try locale detected by jQuery.i18n library
+  // if all else fails, give them English
+  return getConfigForLocale(localeParam) || getConfigForLocale(detectedLocale) || { en: locales.en };
```
drwestco: Fallback {en: locales.en} has a different signature than what getConfigForLocales returns. Might be cleaner to pass an array of locales to geConfigForLocale and have it return the locale+map for the first entry that exists.

public/i18n.js
```diff
@@ -11,10 +11,45 @@ const locales = {
...
+    return null;
+  }
+
+  if (locales[locale]) {
+    return {
+      locale: locale,
+      map: {
+        [locale]: locales[locale]
+      }
+    };
```
drwestco: This might be weird, but the return could just be:

```javascript
return {
  locale: locale,
  [locale]: locales[locale]
};
```

Then you can pass the returned localeConfig directly to both $.i18n() and load()

public/i18n.js
```diff
@@ -11,10 +11,45 @@ const locales = {
...
+
+  const language = locale.split('-')[0];
+
+  if (language && locales[language]) {
+    return {
+      locale: language,
+      map: {
+        [language]: locales[language]
+      }
+    };
```
drwestco: Would be nice to combine the two returns so you're not repeating yourself.

public/i18n.js
```diff
@@ -46,12 +48,10 @@ const determineLocaleConfig = (detectedLocale) => {
...
   // if all else fails, give them English
-  return getConfigForLocale(localeParam)
-    || getConfigForLocale(detectedLocale)
-    || {
-      locale: 'en',
-      map: { en: locales.en }
-    };
+  return getConfigForLocales([localeParam, detectedLocale]) || {
+    locale: 'en',
+    map: { en: locales.en }
```
drwestco: You renamed locales to localesMap. You can just add 'en' to the array.

PR #388, opened 2020-03-31
Added file output to data-{country}.csv in Google storage

marcuscollins:
Added an output to google cloud storage by creating a string in functions/index.js:snapshotData() and writing it out, follow the pattern for the JSON file upload.

I don't currently have a functioning test environment for this step.

### DIFF
--- a/functions/index.js
+++ b/functions/index.js
@@ -193,7 +193,9 @@ async function getSpreadsheet(country, client) {
 }
 
 async function snapshotData(country) {
-  const json_filename = `data-${country}.json`;
+    const base_filename = `data-${country}`;
+    const csv_filename = `${base_filename}.csv`;
+    const json_filename = `${base_filename}.json`;
   const html_snippet_filename = `data_snippet-${country}.html`;
 
   // Talk to sheets.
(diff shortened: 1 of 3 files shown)


functions/index.js
```diff
@@ -250,6 +252,23 @@ async function snapshotData(country) {
     predefinedAcl: "publicRead",
   });
 
+  // Build a simple csv file
+  const csvFileRef = admin.storage().bucket().file(csv_filename);
+  let csv_file_list = [];
+  for (let i = 0; i < data.values.length; i++) {
+      row_string = data.values[i].join(",");
```
drwestco, line 259: You may need to handle quotes, newlines, and commas in the values. "Accepting" can contain commas. "Address" and "Instructions" can contain newlines, quotes, and all sorts of other nonsense.

functions/index.js
```diff
@@ -250,6 +252,24 @@ async function snapshotData(country) {
     predefinedAcl: "publicRead",
   });
 
+  // Build a simple csv file
+  console.log("Preparing CSV File\n");
+  const csvFileRef = admin.storage().bucket().file(csv_filename);
+  let csv_file_list = [];
+  for (let i = 0; i < data.values.length; i++) {
+      row_string = JSON.stringify(data.values[i]);
```
drwestco, line 260: JSON.stringify will use \n and \" for newlines and double quotes. That's not appropriate for csv.

marcuscollins:
> This isn't correct either. JSON has different escaping rules than CSV. To escape a quote in JSON you do `\"`, in CSV you do `""`. There are CSV libraries for Javascript you should be using.

switched to csv-stringify, tested that it escapes as requested. Tested creating csv file. You can download here: https://console.cloud.google.com/storage/browser/_details/findthemasks-dev.appspot.com/data-us.csv?project=findthemasks-dev&folder&organizationId&supportedpurview=project

Opened file in Excel and terminal to confirm formatting.

functions/index.js
marcuscollins: I've switched it--that wasn't my comment.

Also, my job requires lots of parsing CSVs, and I personally detest using a `"` as an escape character. The quote character and the escape character should never be the same!

Anyway, I relent. See new revision.

awong-dev:
sure...merged...but why?  Can't we just do json -> CSV locally in a simple script?

PR #392, opened 2020-04-01
redirect to language specific country donation site forms if available

nburt:


graemer:
Awesome @nburt - do you expect this to test well locally? Any issues there?

Issue #387, opened 2020-03-31
Make entry-approval flow case-insensitive


4kbt:
Right now, entries only go live on the site if the first character in the Approval field is 'x'. Sometimes moderators enter 'X'. Let's make it case-insensitive :)!

jamesarosen:
https://github.com/findthemasks/findthemasks/pull/395/files#diff-1e53ef2f51bd446c876676ba83d7c888R208-R210 addresses this (but isn't merged yet)

4kbt:
I believe that this is fixed.

PR #395, opened 2020-04-01
Make approved column value case insensitive and remove duplicate checks

saikatbhadra:
* Makes both 'x' and 'X' publish a moderated row
* Removes duplicate checks in `toDataByLocation` to check if the row is approved since all entries in data.json should be approved by definition

## Testing
@awong-dev - I don't know how to test this, can you test?

Closes: https://github.com/findthemasks/findthemasks/issues/387

### DIFF
--- a/functions/index.js
+++ b/functions/index.js
@@ -313,7 +313,6 @@ async function getLatLng(address, client) {
 
 function toDataByLocation(data) {
   const headers = data.values[1];
-  const approvedIndex = headers.findIndex( e => e === 'approved' );
   const stateIndex = headers.findIndex( e => e === 'state' );
   const cityIndex = headers.findIndex( e => e === 'city' );
   const data_by_location = {};
--- a/functions/index.js
+++ b/functions/index.js
@@ -24,7 +24,7 @@ const SHEETS = {
 };
 
 if (functions.config().googleapi !== undefined) {
-  CONFIG_CLIENT_ID = functions.config().googleapi.client_id
+  CONFIG_CLIENT_ID = functions.config().googleapi.client_id;
   CONFIG_CLIENT_SECRET = functions.config().googleapi.client_secret;
   GOOGLE_MAPS_API_KEY = functions.config().findthemasks.geocode_key;
 }
--- a/functions/index.js
+++ b/functions/index.js
@@ -90,6 +90,13 @@ async function getAuthorizedClient() {
 
 function splitValues(data) { return [ data.values[0], data.values[1], data.values.slice(2) ]; }
 
+/**
+ * Geocodes entries that are approved in moderation google sheet
+ *
+ * @param data
+ * @param sheet_id
+ * @returns {Promise<void>}
+ */
 async function annotateGeocode(data, sheet_id) {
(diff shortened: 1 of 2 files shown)


functions/index.js
```diff
@@ -292,7 +318,7 @@ function toDataByLocation(data) {
   const cityIndex = headers.findIndex( e => e === 'city' );
   const data_by_location = {};
 
-  const published_entries = data.values.slice(1).filter((entry) => entry[approvedIndex] === "x");
+  const published_entries = data.values.slice(1);
```
awong-dev, line 321: If we remove this filter here, we need to change 279 to take approved_rows...which is probably more correct anyways.

awong-dev: ping! I think this still needs to be resolved.

awong-dev:
ping! Got a conflict here. :-/

functions/index.js
```diff
@@ -24,7 +24,7 @@ const SHEETS = {
 };
 
 if (functions.config().googleapi !== undefined) {
-  CONFIG_CLIENT_ID = functions.config().googleapi.client_id
+  CONFIG_CLIENT_ID = functions.config().googleapi.client_id;
```
jamesarosen, line 27: It looks like the project doesn't have Prettier or anything else installed to ensure standardized code formatting. Are folks interested in that? If so, I'm happy to add it.

functions/index.js
```diff
@@ -237,7 +263,7 @@ async function snapshotData(country) {
     throw new Error("sheet missing expected columns. Ensure row 2 headers are sane?");
   }
 
-  const approved_rows = trimmed_values.filter(e => e[approvedIndex] === "x");
+  const approved_rows = trimmed_values.filter(e => isApproved(e[approvedIndex]));
```
jamesarosen, line 266: 👍

public/toDataByLocation.js
```diff
@@ -1,11 +1,10 @@
 export default function toDataByLocation(data) {
   const headers = data.values[1];
-  const approvedIndex = headers.findIndex( e => e === 'approved' );
   const stateIndex = headers.findIndex( e => e === 'state' );
   const cityIndex = headers.findIndex( e => e === 'city' );
   const data_by_location = {};
 
-  const published_entries = data.values.slice(1).filter((entry) => entry[approvedIndex] === "x");
+  const published_entries = data.values.slice(1);
```
jamesarosen, line 7: An alternative way to write this (not better or worse) would be
```js
const [, ...published_entries] = data.values;
```

functions/index.js
awong-dev: We do use eslint inside the cloud functions.

I had been initially resistant to adding formatting tools since I expected this project to sunset in ~2 weeks. However, we're still around. :)

If you want to do something to standardize our formatting, I'd be happy to approve! I'm used to eslint, but honestly have very little opinions on the specific tool.  However, it'd be nice to use the same tool everywhere.

awong-dev:
This is super stale so closing for now. Feel free to reopen if we still want this !

Issue #378, opened 2020-03-31
Make the site faster by rendering fewer things


raindrift:
I'm getting too tired tonight to simply implement this, but here is a detailed plan. The idea is to stop rendering the complete list of 2k+ sites, and instead render the 50 at the top of the list, and then add more as the list scrolls. This should considerably improve pageload time (which is currently ~50% rendering, even on a recent macbook pro), and I have also verified that it'll make the google map usable again on a middle-aged Android phone (Pixel 2).

The plan:

1. Render the first 50 elements in the chosen set, and then render more in sets of 50 with infinite scroll. Doing infinite scrolling well is hard, so ideally we'll use a plugin for that. I think the one at https://infinite-scroll.com will work, but really anything is fine as long as it can be convinced not to load anything from an ajax endpoint, but instead just send an event when the window scrolls to the right position.

The best way to do this would be:
 - getFilteredContent returns only the entries for the content to be displayed, instead of the DOM elements (currently grouped into `<div>`s by state)
 - the output of getFilteredContent is stored in a convenient variable
 - the first 50 elements are rendered into the page immediately with a new renderContent function
 - the offset is stored
 - when the page is scrolled, the next 50 entries are rendered, etc.
 - when the filters change, the offset and content are replaced, and the content div is emptied
 - the renderContent function will need to be responsible for displaying a new state header when necessary, and the DOM will need to be reorganized a little so that a partial state can be displayed without breaking the formatting.

2. (nice to have) It would be more performant to stop generating DOM for every entry at pageload (the `createContent` function). Instead, build the in-memory representation of the DOM as it's needed. Ideally this means creating a function to return the generated DOM for each entry and memoizing its output (either in the window.data_by_location object directly, or in a cache object keyed by entry id). Then, renderContent would call this function to get/generate the DOM for a given entry.

More in this thread, including links to caveats about infinite scrolling: https://findthemasks.slack.com/archives/C010J52HXPD/p1585632813145300

drwestco:
WIP for list performance: https://github.com/findthemasks/findthemasks/tree/LocationsListPerf

List starts out empty until one or more states is selected. Map view shows all locations, but dims those that are outside the selected states. Needs some polish (counter doesn't do the right thing, interaction with search bar, etc.), but could be a useful starting point.

PR #396, opened 2020-04-01
Render the list 40 entries at a time with infinite scroll.

raindrift:
This should improve pageload times a great deal, and fix Android map performance issues. Implements both parts 1 and 2 of #378.

Page should look and work the same. Here are the before and after from the profiler:
Screenshot 2020-03-31 20 32 18
Screenshot 2020-03-31 20 31 43

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -231,17 +231,7 @@
         white-space: nowrap;
       }
 
-      .state {
-        display: flex;
-        flex-wrap: wrap;
-        position: relative;
-      }
-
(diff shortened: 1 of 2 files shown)


nburt:
Great work, this is so much faster.

I am seeing a bug when you hit the last results where it keeps appending the last result's city to the dom. Here's with state filters applied:

image

Here's in CA:
image

patricknelson:
This is such a vast improvement; awesome work! 🚀

PR #397, opened 2020-04-01
add Portugal, Poland, Spain, Switzerland, UK, Austria, Italy, Germany

nburt:
Before merge make sure the following checklist is complete:

Portugal
- [x] Missing symlink
- [x] Manually correct MT translation: Precinct -> Distrito
- [x] Merge config

Poland
- [ ] forms are a copy of Switzerland, including Canton as administrative region. This needs to be updated. Then, we'll need another PR with the updated region name. Currently it is using the string for US State.
- [x] Missing symlink
- [ ] Merge config

Spain
- [x] Missing symlink
- [x] Merge config

Switzerland
- [x] Merge config

UK
- [ ] test that filters work for UK which doesn't have an administrative region (US state, CA province, FR department, etc.) for its form
- [x] Merge config

Austria
- [x] Manually correct MT translation: Federal State -> Bundesland (Nate)
- [x] Missing symlink
- [x] Merge config

Italy
- [x] Missing symlink
- [x] Merge config

Germany
- [x] Manually correct MT translation: State -> Land (Nate)
- [x] Confirm that German states should be called "Land" rather than "Bundesland" as in Austria (Nate)
- [x] Merge config

France
- [x] Merge config

For all of the above we should also be testing that the data loads, language navigation, form navigation, maps, and filtering.


awong-dev:
wait, it says before merge, ensure the checlist is done. I haen't odne that. :-/

nburt:
@awong-dev just commented out UK / Poland. I'll follow up on the Poland form + UK state filter tomorrow

public/countries.js
```diff
@@ -39,6 +50,84 @@ export default {
...
+    i18nString: 'ftm-portugal',
+    countryCode: 'pt',
+    administrativeRegionI18nString: 'ftm-administrative-region-precinct',
+    donationSites: {
+      i18nString: 'ftm-poland-large-donations',
+      administrativeRegionLinks: null,
+      nationalLinks: null
+    },
+    noDonationSitesNearMeI18nString: 'ftm-poland-no-donation-sites-near-me'
+  },
```
awong-dev, line 97: Can we put country clauses in alphabetical order please?

nburt: yeah so these are actually put in alphabetical order by English name. I'll update to country code and then do some sorting

PR #400, opened 2020-04-01
Filter map by state

drwestco:
Extension of the show-others feature. Instead of hiding map items outside the currently-selected state, dim them, as we do for data in other countries. Also change the dot pin icon to svg, since the built-in circle icon is quite slow to create.

Map filtered to KY + WV. Entries from surrounding states still show, but dimmed.
image


public/locations-list-map.js
```diff
@@ -726,11 +730,11 @@ function getMarkers(data, appliedFilters, bounds, markerOptions) {
         let marker = entry.marker;
 
         if (marker) {
-          if (!inStateFilter || !inAcceptFilter) {
+          if (!inAcceptFilter) {
```
drwestco, line 733: Logic here is a little weird if there's an "accepts items" filter set. We'll filter the in-state markers (those that don't match the "accepts items" filter will be hidden). But we'll leave the out-of-state markers alone. That's consistent with the show-others features - pins for all out-of-country entries are all shown.

drwestco: Ah, this weird logic breaks behavior when no state filter is set. All items are then considered 'out of state' and the "accepts items" filter has no effect.

nburt:
One thing I noticed is that when clearing the state filter the clustering seems a bit wonky:

image

Once I zoom in or out after it resets itself back to normal.

I also find the scroll on state select pretty annoying. Maybe for another conversation but given that we're zooming in on states I'd rather navigate donation sites that way.

Another separate question, why doesn't the show-others filter by country apply to the US?

drwestco:
Ooh, I haven't seen the clusters freak out like that (though with the auto-scrolling behavior, maybe I just haven't noticed).

show-others was originally for non-US markets only since the desire was to make up for lack of preexisting data when bootstrapping additional countries. If this is fast enough and a good enough experience, we can certainly use this for all.

PR #405, opened 2020-04-01
Scroll to map on filter change

drwestco:
We've gone back and forth on whether changing a filter should scroll to the list, to the map, or not at all. How about a solution where we scroll to the map if it's available, or to the list if the page was created with ?hide-map=true? I'm probably biased since I want to show off the show-others feature, so I'm open to feedback on this behavior...


graemer:
I've gotten a number of points of feedback that the autoscroll behavior is jarring. My bias at this point is simply to remove. @drwestco thoughts?

drwestco:
Removed auto-scroll in latest iteration. May need to revisit for mobile since the filters section is so large.

PR #407, opened 2020-04-01
Feature/google analytics user interactions

### LINKED ISSUE #373: Fire Google Analytics events for key user interactions
Generate an event
- [x] if a healthcare worker clicks on the link to add/remove/update a donation site
- [x] if a user navigates to the large donation link
- [x] if a user navigates to the general getusppe.org link
- [x] if a user goes to our Facebook page
- [x] if a user clicks on a map cluster
- [x] if a user clicks on a map marker

ngottlieb:
resolves #373

### DIFF
--- a/public/locations-list-map.js
+++ b/public/locations-list-map.js
@@ -104,6 +104,14 @@ const addDonationSites = () => {
 
     largeDonationElement.innerHTML = $.i18n(countryDonationSites.i18nString, administrativeRegionStringHtml, nationalLinksHtml);
     noDonationsElement.innerHTML = $.i18n(countryConfig.noDonationSitesNearMeI18nString);
+
+    $(largeDonationElement).find('a').click(function(e) {
+      sendEvent('largeDonation', 'click', $(this).attr('href'));
+    });
+
+    $(noDonationsElement).find('a').click(function(e) {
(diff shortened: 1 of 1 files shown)


public/locations-list-map.js
```diff
@@ -104,6 +104,14 @@ const addDonationSites = () => {
 
     largeDonationElement.innerHTML = $.i18n(countryDonationSites.i18nString, administrativeRegionStringHtml, nationalLinksHtml);
     noDonationsElement.innerHTML = $.i18n(countryConfig.noDonationSitesNearMeI18nString);
+
+    $(largeDonationElement).find('a').click(function(e) {
+      sendEvent('largeDonation', 'click', $(this).attr('href'));
```
saikatbhadra, line 109: I assume click handlers still work when you redirect the user to an external url?

saikatbhadra:
Looks good to me, @graemer can you merge ?

public/locations-list-map.js
ngottlieb: it does...although thinking about that question raises another one that I don't know the answer to -- is that guaranteed by spec, or is it possible a browser implementation (or weird set of circumstances) would lead to it not executing the new click event? not really anything to worry about here but I am going to look into it.

PR #410, opened 2020-04-01
City ordering fix, removed scrollbar, styles on give.html

raindrift:
Cities are now listed in order again, random horizontal scrollbar removed

### DIFF
--- a/public/locations-list-map.js
+++ b/public/locations-list-map.js
@@ -264,11 +264,13 @@ function getFlatFilteredEntries(data, filters) {
     for (const cityName of Object.keys(cities).sort()) {
       const city = cities[cityName];
 
-      for (const entry of city.entries) {
+      city.entries.sort(function (a, b) {
+          return a.name.localeCompare( b.name );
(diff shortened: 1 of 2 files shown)


public/locations-list-map.js
```diff
@@ -264,19 +264,21 @@ function getFlatFilteredEntries(data, filters) {
     for (const cityName of Object.keys(cities).sort()) {
       const city = cities[cityName];
 
-      for (const entry of city.entries) {
+      city.entries.sort(function (a, b) {
+          return a.name.localeCompare( b.name );
+      }).forEach(function(entry) {
```
drwestco, line 269: Why switch from for..of to forEach?

raindrift: It seemed clearer to use chaining than to cram the entire comparator into the args for `for`, but I'm not really attached to any particular syntax. If you care a lot about this, I'm happy to change it.

drwestco: nah, I was just curious

PR #413, opened 2020-04-02
Move styles to shared css file

drwestco:
Move styles for index and give pages to shared styles.css. Add a .give class at the container for a couple of styles that need to vary between the two pages (full width map vs partial width, etc.)


awong-dev:
FYI, branch has conflict now :(

PR #422, opened 2020-04-02
add cors header for new country data.json files

ngottlieb:


ngottlieb:
resolves #423

alex:
I'm pretty sure Apache supports a file regexp for these things, which would save the trouble and duplication.

drwestco:
Regex suggestion is a good one. https://www.askapache.com/htaccess/using-filesmatch-and-files-in-htaccess/

ngottlieb:
copy that, was blocking what I was working on (slightly) so I just mimicked what was already there but I will update this PR with regex instead

ngottlieb:
@drwestco updated

PR #442, opened 2020-04-04
thanks for corporate donations, add headers to whoweare page

nburt:


awong-dev:
Merging for now @r-pop can you double check to see it looks okay to you?

PR #447, opened 2020-04-04
Fixes #446: translate enum values from forms

### LINKED ISSUE #446: Translate PPE in the "Accepting" section in marker popups and the locations list
We now have a Merge Config for form submissions in Google Sheets that will "translate" enums from languages -> English so that we can correctly apply filters.

Since these enums are saved in the `data-{countryCode}.json` files in English, they are rendered in marker popups and the locations list in English, regardless of the locale you select.

We need to look up these enum keys dynamically and render them in the selected locale.

nburt:
I'm open to some other solutions here, it's a little brittle to be hardcoding these values in case they change but seems like an OK v1.

I'm thinking that a 2nd iteration could be to parse en.json during deploys and write a JS file with all the keys in en.json reversed. This would replace `enumKeyLookups.js` on each deploy


public/i18n/enumKeyLookups.js
```diff
@@ -0,0 +1,20 @@
+export default {
```
drwestco, line 1: What's the difference between the *item* values and *accepting*?

drwestco, line 1: Need to update filters.acceptItems to match these new values? i.e., 'overalls' will match both Medical overalls and Coveralls/bunny suits. 'booties' will match both Medical and Disposable booties. Masks will match both Surgical and Homemade. etc. etc.

nburt: Item was an existing string and seemed to be used for filters. When I added missing strings I namespaced with "accepting" because they're from the form field called "accepting". I'll take another pass and make naming more consisten

nburt: hmm ok. I think I need to talk to someone a bit about the filters. I'll post in data, I think they've been updating the forms and it's worth a revisit on what the filters should be and how they should match

public/locations-list-map.js
```diff
@@ -148,6 +150,23 @@ const addDonationSites = () => {
...
+const translatedFilterItems = () => {
+  const translated = {};
+  const filterItemKeys = Object.keys(FILTER_ITEMS);
+
+  for (let i = 0; i < filterItemKeys.length; i++) {
+    const filterItemKey = filterItemKeys[i];
+    const filterItem = FILTER_ITEMS[filterItemKey];
+
+    translated[filterItemKey] = {
+      name: $.i18n(filterItem.name),
```
nburt, line 162: needed to wait until i18n has loaded before calling which is why these can't live in the other file

public/i18n/enumKeyLookups.js
crayne: I'm wondering about the open packages choices. You have yes and no, but what about "other" on the form.  Or should there be an "other" on the form?

public/locations-list-map.js
awong-dev: This should be a code comment...

public/locations-list-map.js
```diff
@@ -148,6 +150,23 @@ const addDonationSites = () => {
   }
 };
 
+const translatedFilterItems = () => {
+  const translated = {};
+  const filterItemKeys = Object.keys(FILTER_ITEMS);
+
+  for (let i = 0; i < filterItemKeys.length; i++) {
```
awong-dev, line 157: nit: `for (const [filterItemKey, filterItem] of Object.entries(FILTER_ITEMS)` is the same thing

public/locations-list-map.js
```diff
@@ -982,6 +990,26 @@ function centerMapToBounds(map, bounds, maxZoom) {
...
+  }
+
+  return value;
+};
+
+const translateEnumList = (enumListString) => {
+  if (enumListString) {
+    return enumListString.split(', ').map((stringValue) => (
+      translateEnumValue(stringValue)
+    )).join(', ')
```
awong-dev, line 1007: I'd tend to `split(',').map(s => translateEnumValue(s.trim()))` to make it robust to whitespace mistakes. But this is fine too.

public/i18n/enumKeyLookups.js
nburt: There is an "other" category for both the "items accepted" and the "open packages" choices however I think it is less important to translate those.

This is the current Google Forms process:

1. User submits form in some language (e.g. French)
2. AppScript converts the "enum" categories to English so that filtering can be applied in a single language
3. Sheet written to data.json file
4. App loads sheet and renders it

What we're doing here is translating all the "enum" categories that we know are options. We could do the other columns but we'd need to use something like Google Translate so for now I think it's best to just leave the freeform fields as is

PR #449, opened 2020-04-04
improve map geocoder search usability by zooming to recommended viewport

### LINKED ISSUE #436: Map-zoom on search is too tight
When searching for a state, the Map view, at least sometimes, zooms in too tightly.

When searching for "California", it picks out a few pins near Fresno. "Texas" favors a few points near Dallas. This one might be tricky to fix.

image

ngottlieb:
resolves #436

### DIFF
--- a/public/locations-list-map.js
+++ b/public/locations-list-map.js
@@ -680,9 +680,9 @@ function initMapSearch(data, filters) {
     if (place.geometry) {
       // Get the location object that we can map.setCenter() on
       sendEvent("map","autocomplete", $search.val());
-      let location = place.geometry.location;
-      if (location) {
-        centerMapToMarkersNearCoords(location.lat(), location.lng())
+      let viewport = place.geometry.viewport;
+      if (viewport) {
+        fitMapToMarkersNearBounds(viewport);
       } else {
         sendEvent("map","autocomplete-fail", $search.val());
         console.warn('Location data not found in place geometry (place.geometry.location).')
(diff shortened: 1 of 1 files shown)


public/locations-list-map.js
```diff
@@ -770,34 +787,37 @@ function centerMapToMarkersNearCoords(latitude, longitude) {
...
+  // order markerDistances by key (distance)
+  let distances = [...markerDistances.keys()].sort((a,b) => a -b);
+  // return array of markers in order of distance ascending
+  return distances.map((distance) => markerDistances.get(distance));
+}
+
+/**
+ * Centers map around markers nearest to an arbitrary set of latitude/longitude coordinates.
+ */
+function centerMapToMarkersNearCoords(latitude, longitude) {
```
drwestco, line 802: You should be able to get rid of this function and just use fitMapToMarkersNearBounds. In the "near me" case, you can construct a bounds with the user's location. The hasMarker logic should apply to fitMapToMarkersNearBounds too.

public/locations-list-map.js
```diff
@@ -752,14 +752,31 @@ function centerMapToMarkersNearUser() {
 }
 
 /**
- * Centers map around markers nearest to an arbitrary set of latitude/longitude coordinates.
+ * Fits map to bounds, expanding the bounds to include at least three markers as necessary.
+*/
+function fitMapToMarkersNearBounds(bounds) {
+  // get center of bounding box and use it to sort markers by distance
+  let center = bounds.getCenter();
+  const markersByDistance = getMarkersByDistanceFrom(center.lat(), center.lng());
```
drwestco, line 760: Since you know you only need 3, you can pass that to getMarkersByDistanceFrom.

public/locations-list-map.js
```diff
@@ -770,34 +787,37 @@ function centerMapToMarkersNearCoords(latitude, longitude) {
...
-    distances.push(distance);
+    markerDistances.set(distance, marker);
   }
 
-  // sort the distances and set bounds to closest three
-  distances.sort((a, b) => a - b);
+  // order markerDistances by key (distance)
+  let distances = [...markerDistances.keys()].sort((a,b) => a -b);
+  // return array of markers in order of distance ascending
+  return distances.map((distance) => markerDistances.get(distance));
```
drwestco, line 796: If getMarkersByDistanceFrom takes the desired length as a param, you can slice the top n off the array and just return those markers instead of all of them.

nburt:
shoot just merged, @ngottlieb can you follow up with @drwestco suggestions?

ngottlieb:
copy that, I'll make the changes and submit a new PR @drwestco @nburt

PR #450, opened 2020-04-04
Add google maps link to address section of marker popover

saikatbhadra:
Screen Shot 2020-04-04 at 3 06 06 PM

### DIFF
--- a/public/locations-list-map.js
+++ b/public/locations-list-map.js
@@ -490,6 +490,14 @@ function renderNextListPage() {
   lastLocationRendered = renderLocation - 1;
 }
 
+function getOneLineAddress(address) {
+  return address.trim().replace(/\n/g, " ");
+}
+
+function googleMapsUri(address) {
+  return encodeURI(`https://www.google.com/maps/search/?api=1&query=${address}`);
+}
+
 function getEntryEl(entry) {
   if (!entry.domElem) {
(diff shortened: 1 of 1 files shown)


linderr:
@saikatbhadra does your code work with Google Maps plus codes?? Trying to find a work around when several departments in a hospital request. They have different needs and instructions sometimes. For example pinning an Emergency Room bay vs. main hospital lobby

nburt:
@linderr can you provide some examples?

My 5 minutes of googling understanding is that this is possible but it won't necessarily work out of the box and we might need to decode the plus code. Here's the JS implementation: https://github.com/google/open-location-code/tree/master/js

linderr:
@nburt @saikatbhadra I found a manual work around in the Sheets --- b/c Combined is just a output that can be manually over-written, I use the plus codes in the Final Address column of the spreadsheet to trick the lat/lng bot to give me the EXACT geocode of the **portion** of the hospital I want. After it comes by and spits that out, I replace the Final Address with the front-facing street address. Used this on a few hospitals already to distinguish ER department from hospital Foundation (main lobby/administrative building) on the same campus/street address
it's been working well so no need to engineer this for now

linderr:
If either of you are interested we can jump on a call later --- gotta prepare for the CA national call now so i'll be MIA for a few hours :)

PR #458, opened 2020-04-05
add meta tags and parse fb_locale for social sharing translation

nburt:
locale is still the default, we'll fallback to fb_locale if locale is not present. 

based off: https://stackoverflow.com/questions/13923944/how-can-i-internationalize-meta-og-tags-to-be-parsed-correctly-by-facebook


nburt:
Alright, this isn't gonna work because FB doesn't interpret JS. We'd need to switch to server side rendering if we want this to work...

PR #461, opened 2020-04-05
Feature/improve map search zooming usability

ngottlieb:
here's the additional changes requested re: pull request #449.

I'm not totally sure I understand what the `hasMarker` functionality actually does -- what is the scenario where we are fitting bounds (either zooming to the user or to a search result) and don't have a marker?

### DIFF
--- a/public/locations-list-map.js
+++ b/public/locations-list-map.js
@@ -792,8 +792,6 @@ function fitMapToMarkersNearBounds(bounds) {
 
   const markersByDistance = getMarkersByDistanceFrom(center.lat(), center.lng(), 3);
 
-  let hasMarker = false;
-
   // extend bounds to fit closest three markers
   markersByDistance.forEach((marker) => {
       bounds.extend(marker.position);
--- a/public/locations-list-map.js
+++ b/public/locations-list-map.js
@@ -795,15 +795,11 @@ function fitMapToMarkersNearBounds(bounds) {
   let hasMarker = false;
 
   // extend bounds to fit closest three markers
-  [0,1,2].forEach((i) => {
-    const marker = markersByDistance[i];
-    if (marker) {
-      hasMarker = true;
+  markersByDistance.forEach((marker) => {
       bounds.extend(marker.position);
-    }
   });
 
-  if (hasMarker) {
+  if (!bounds.getNorthEast().equals(bounds.getSouthWest())) {
     // zoom to fit user loc + nearest markers
     map.fitBounds(bounds);
   } else {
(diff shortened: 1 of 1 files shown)


drwestco:
"what is the scenario where we are fitting bounds (either zooming to the user or to a search result) and don't have a marker?"
It was pretty common before data started filling in for other countries. Visitors still can also set some combination of filters that exclude all markers. It's really just a catch-all so we don't zoom the map in to level 22 and lose our place.

public/locations-list-map.js
```diff
@@ -768,23 +770,32 @@ function centerMapToMarkersNearUser() {
...
   // get center of bounding box and use it to sort markers by distance
   let center = bounds.getCenter();
-  const markersByDistance = getMarkersByDistanceFrom(center.lat(), center.lng());
-  
+  const markersByDistance = getMarkersByDistanceFrom(center.lat(), center.lng(), 3);
+
+  let hasMarker = false;
+
   // extend bounds to fit closest three markers
   [0,1,2].forEach((i) => {
```
drwestco, line 778: Now you can just do markersByDistance.forEach, and don't need the if(marker) test.

public/locations-list-map.js
```diff
@@ -768,23 +770,32 @@ function centerMapToMarkersNearUser() {
...
     const marker = markersByDistance[i];
     if (marker) {
+      hasMarker = true;
       bounds.extend(marker.position);
     }
   });
 
-  map.fitBounds(bounds);
+  if (hasMarker) {
```
drwestco, line 786: I wouldn't do this. I'd check bounds.getNorthEast and bounds.getSouthWest and see if they're different. If so, call fitBounds. If not, call setCenter.

public/locations-list-map.js
```diff
@@ -804,38 +815,7 @@ function getMarkersByDistanceFrom(latitude, longitude) {
...
-  });
-  
-  if (hasMarker) {
-    // zoom to fit user loc + nearest markers
-    map.fitBounds(bounds);
-  } else {
-    // just has user loc - shift view without zooming
-    map.setCenter(latlng);
-  }
+  return distances.map((distance) => markerDistances.get(distance)).slice(0, n);
```
drwestco, line 818: Flip it around. distances.slice(0, n).map(...)

ngottlieb:
@drwestco updated. As far as I can tell, the former `hasMarker` logic (and now, "bounds is a point" logic) will only ever be used when pinning to user location. It's possible Geocoding could return a point as "viewport" in some situation but I don't know what that situation would be...

awong-dev:
please resolve conflict and I'll merge. :)

saikatbhadra:
FYI @ngottlieb looks like you have a merge conflict :(

ngottlieb:
Sorry, somehow missed that original comment, thanks for heads up @saikatbhadra. Should be good to go now.

PR #467, opened 2020-04-05
Updating copy below map for Canada only

sumitsekhri:
I just updated the copy below for the Give page. Not sure if I've modified the correct file but happy to make changes if not.


nburt:
@sumitsekhri @linderr This updates fallback for all countries rather than just Canada

linderr:
@nburt sorry, we're asking Ewa to walk us through the pipeline from edits->Smartling->more

PR #469, opened 2020-04-05
WIPR: Get maps API key from env var that can be set in heroku

maplebird:
WIP do not review yet


awong-dev:
edited title to WIPR: to indicate work in progress.

PR #482, opened 2020-04-06
Fixes #480: adjust styles to better handle longer languages on mobile

### LINKED ISSUE #480: French translations overflow on mobile
There are a few places where we must be using fixed pixel widths and longer languages like French (~20% longer than English) overflow and are cutoff. We should fix this instance and check others and adjust our styles. Found this my Samsung J7 Prime.

Screenshot_20200406-084442_Brave

nburt:


public/index.html
```diff
@@ -85,7 +85,7 @@
 
         </div>
         <div class="subhead">
-          <div class="subhead-left i18n" data-i18n="ftm-index-need-helpers">Let's Help The Helpers</div>
+          <div class="subhead-left i18n mr-lg-3 py-4 px-3" data-i18n="ftm-index-need-helpers">Let's Help The Helpers</div>
```
drwestco, line 88: Why do you need the additional classes? Can't you just add the relevant margins+padding to the .subhead-left rule?

drwestco: The .py-4 padding top/bottom is pretty hefty on mobile.

image

nburt: We don't really. What are your thoughts here? I'm thinking we may as well just try and lean into bootstrap styling where possible since it's already defined which is why I switched to their padding styles. I'll make a tweak to the padding on mobile and merge but can follow up tonight

PR #484, opened 2020-04-06
gracefully fallback to english if locale not found

nburt:


nburt:
deploying as hotfix, czech user is having issues

PR #494, opened 2020-04-07
create filters using the field translations config

nburt:


nburt:
@drwestco yeah the forms are out of sync. Step 1 I'm gonna try and work with the data teams in FR / CA and get the forms into sync so we use the same terms. If they need to be country specific I'll translate them.

public/formEnumLookups.js
```diff
@@ -62,6 +62,17 @@ export const FILTER_ITEMS = {
   'nasopharyngeal swabs': {
     name: 'ftm-accept-item-nasopharyngeal-swabs',
     isSet: false
+  },
+  // France uses certain country specific naming for accepted items
+  'ffp2 masks (n95) or ffp3': {
+    name: 'ftm-accept-item-france-n95s',
+    isSet: false,
+    countrySpecific: 'fr'
```
nburt, line 70: this is European countries, not just FR

nburt: @drwestco any thoughts on the best way to handle this? At this point every country except for the US / CA would have this item on their form since in the EU N95 masks are called FFP2. Maybe a `countryBlacklist` array?

nburt: Now that I think about it, as soon as we can migrate the US to use Merge Config this is a moot point since I believe it is the last holdout not using it. @awong-dev do you think this would be possible to do today?

nburt:
going with country blacklist to start, can follow up if necessary

Issue #377, opened 2020-03-31
Display a country specific banner


saikatbhadra:
From #dev:
> As we expand into more countries in the world, is there anyway to display a country-specific banner? Right now it's unclear which country you're looking at until you get to the map/list. Is there some universal map icon set we can use?
> 
> Example:

![image](https://user-images.githubusercontent.com/2250961/77991568-21471e00-72d9-11ea-9a71-265f0a910a61.png)

PR #500, opened 2020-04-07
Display country banner flag

ywyw:
Addresses https://github.com/findthemasks/findthemasks/issues/377

Here's an example of what it looks like:

Screen Shot 2020-04-07 at 12 27 18 AM

Screen Shot 2020-04-07 at 12 36 05 AM
Screen Shot 2020-04-07 at 12 36 22 AM
Screen Shot 2020-04-07 at 12 36 29 AM


functions/package.json
```diff
@@ -14,11 +14,12 @@
   },
   "dependencies": {
     "@googlemaps/google-maps-services-js": "^2.3.1",
+    "csv-stringify": "",
     "firebase-admin": "^8.6.0",
     "firebase-functions": "^3.3.0",
+    "flag-icon-css": "^3.4.6",
```
drwestco, line 20: Not sure why changes to package and package-lock are necessary, since you're linking directly to the CDN-hosted css.

public/flag.js
```diff
@@ -0,0 +1,5 @@
+$(window).on( "load", function() {
```
drwestco, line 1: You can add this to locations-list-map, instead of creating a new JS file. There's already logic to get the current country code in that file.

public/flag.js
```diff
@@ -0,0 +1,5 @@
...
+  // get url path and strip leading/trailing slashes, if no path, assume US
+  var countryCode = window.location.pathname.replace(/^\/|\/$/g, '') || 'us';
+  $('#flag-icon-span').removeClass().addClass('flag-icon flag-icon-' + countryCode);
```
drwestco, line 4: Don't need to removeClass, just add the specific flag-icon-xx class.

ywyw:
Updated screenshots from using https://github.com/hjnilsson/country-flags:

Screen Shot 2020-04-07 at 10 50 59 AM
Screen Shot 2020-04-07 at 10 51 18 AM
Screen Shot 2020-04-07 at 10 51 33 AM
Screen Shot 2020-04-07 at 10 52 00 AM

public/style.css
```diff
@@ -73,6 +73,22 @@ hr {margin-top: 1.7rem;}
...
+.flag-wrapper > img {
+   display: inline-block;
+   margin-left: 90px;
+}
+
+#flag-image {
+  margin-top: 5px;
+  width: 40px;
+}
+
```
drwestco, line 91: Can you simplify these? "#flag-image" and ".flag-wrapper > img" refer to the same element. You can also combine margin-top with .flag-wrapper - currently you're subtracting 35 then adding back 5. Could just subtract 30 on the wrapper and call it done. text-align and display properties don't appear to do anything here either.

ywyw: Oof this was cruft leftover from the old implementation.  You're right, it didn't need most of these

saikatbhadra:
Awesome work!

PR #508, opened 2020-04-07
Share links

crayne:
It appears that we don't have an entry for LinkedIn, so the LindedIn entry does nothing.  For reddit, the link  a search for "findthemasks"


public/index.html
```diff
@@ -26,6 +26,7 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <link href="https://fonts.googleapis.com/css?family=Roboto:400,700&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
```
nburt, line 29: @crayne we had actually decided to remove font awesome last week to improve page load since we were only using it for a Facebook icon. Can you find minified svg or jpg images and use those instead?

public/index.html
```diff
@@ -201,11 +196,16 @@ <h2 class="i18n" data-i18n= "ftm-map-of-sites">Donation Sites</h2>
...
           </p>
-
           <p class="i18n" data-i18n="[html]ftm-made-with-love">made with &lt;3 in Seattle</p>
           <a href="javascript:;" onclick="window.location='whoweare.html'+window.location.search;" class="end i18n" data-i18n="ftm-who-we-are">who we are</a>
-        </div>
-      </div>
+          <div class="social-media">
+            <a href="http://facebook.com/findthemasks" class="fa fa-facebook social-media-icons" target="_blank"></a>
+            <a href="http://twitter.com/findthemasks" class="fa fa-twitter social-media-icons" target="_blank"></a>
+            <a href="http:#" class="fa fa-linkedin social-media-icons" target="_blank"></a>
```
nburt, line 204: I guess if we don't have linkedin we should just remove it

public/index.html
crayne: Page load is much faster now since chunked loading was implemented.  So is it still necessary to remove font awesome?

public/index.html
crayne: Will do.  Are there any social media icons that you think are missing?

public/index.html
```diff
@@ -53,12 +54,6 @@
   </head>
 
   <body>
-    <div class="languages-loading" style="position: fixed;top: 50%; left: 50%;">
```
nburt, line 56: why was this removed?

public/index.html
nburt: Can you replace linkedin with instagram? https://www.instagram.com/findthemasks/

public/index.html
nburt: what's the size diff of font awesome vs. small icons for each social platform? I'd vote for whatever is smallest since we aren't using font-awesome anywhere else at this point

saikatbhadra:
Just posting what I had on the main issue.

> @susanashlock - I am not sure I did a good job of explaining this feature so what I am looking for are buttons where a user clicks and it generates a message for them to post to their network.

So user clicks on Facebook link, it opens a new tab and they get a link to Facebook with a filled out post to their timeline. Something like "Find out where you can donate PPE at findthemasks.com" (Talk to #comms for a better message). The user would be able to edit that message and then post to their timeline. Same goes for LinkedIn, Instagram, whatever networks the cool folks are into nowadays.

I don't want these links to go to our social pages because that doesn't drive traffic to our main site.

Oh and ideally we should have google analytics linked to the click of these buttons, but its definitely just a nice to have (not a blocker).

public/index.html
crayne: Individual icons are much smaller so I'll use them

public/index.html
crayne: not on purpose -- I'll  fix it

crayne:
According the Facebook, prefilling a message is not allowed.  https://developers.facebook.com/docs/apps/review/prefill

saikatbhadra:
Ok let's not prefill. It's better for the user to give their own message anyhow.

crayne:
I will delete utils.js -- it is not needed

crayne:
@nburt Did you request a change for this push?

nburt:
@crayne Looks good to me.

I have 3 questions / comments (not necessarily blocking merge):

1. The diff on index.html is pretty substantial, looks like IDE formatting. Is that possible to revert to only the social changes? If that's a challenge it's ok, I can handle merging these changes with the Node PR which I think is the main one this would conflict with right now.

2. Kind of a separate issue but with the footer where it is these links won't really appear except for countries that don't have many locations. This should be solved by new designs in #ux-design but currently social icons aren't in the mockups, just posted there to make sure they're added.

3. Do you think it'd be better to have Twitter link to our profile or prefill a message to the Tweet? I was a little confused when I ended up here: https://twitter.com/intent/tweet. (Full disclosure I don't really use Twitter so this could just be me)

crayne:
1. This is weird but when I pulled convert-to-nodejs to review it, I saw that the social media button code was integrated there (I must have sleep-coded). If this is the case then maybe I should wait until the node PR is merged and then pull from master to make any further changes.

2.  I can fix this pursuant to the above but maybe not, because of the redesign? Ironically I was filtering by a state with few entries in order to be able to scroll fast to the bottom to test my code.

3.  There was a discussion about this above. @susanashlock favors a pre-filled message, but Facebook doesn't allow it, so I was trying to be consistent.  But if it is confusing then pre-filling makes sense.

public/index.html
```diff
@@ -1,214 +1,237 @@
...
-              </a>
-              <div class="dropdown-menu dropdown-menu-right" aria-labelledby="countries-dropdown" id="countries-dropdown-selector">
-              </div>
-            </li>
-          </ul>
-        </div>
-      </nav>
-        <div class="heading-container">
-          <img class="logo" src="images/logo-color.svg" alt="Find The Masks logo" />
-          <div class="flag-wrapper"><img id="flag-image" alt="Find The Masks flag" /></div>
```
drwestco, line 85: Flag icon should stay.

public/index.html
```diff
@@ -1,214 +1,237 @@
...
+      <div class="question i18n" data-i18n="ftm-comments-questions">Got a comment or suggestion not related to data updates? Want to help?</div>
+      <p class="i18n" data-i18n="[html]ftm-contact-info">
+        Email us at contact@findthemasks.com or join us on <a target="_blank" rel="noreferrer noopener" href="https://github.com/r-pop/findthemasks">our Github</a>.
+      </p>
+      <p class="i18n" data-i18n="[html]ftm-made-with-love">made with &lt;3 in Seattle</p>
+      <a href="javascript:;" onclick="window.location='whoweare.html'+window.location.search;" class="end i18n" data-i18n="ftm-who-we-are">who we are</a>
+      <div>
+              <div class="social-media-icons">
+                <a class="fb-xfbml-parse-ignore" target="popup" onclick="window.open('https://www.facebook.com/sharer/sharer.php?u=http%3A%2F%2Fwww.findthemasks.com%2F&amp;src=sdkpreparse','popup', 'width=600,height=800');return false;" href="https://www.facebook.com/sharer/sharer.php?u=http%3A%2F%2Fdengdeney.blogspot.com.tr%2F&amp;src=sdkpreparse">
```
drwestco, line 210: What is the blogspot URL in the href for?

public/index.html
```diff
@@ -1,214 +1,237 @@
...
+                  <img  src="images/facebook.svg" border="0" />
+                </a>
               </div>
-              <!-- BETA: To enable search field, either change data-enable-search to "true" in #map below, or add hide-search=false/show-search=true to query string. -->
-              <div class="search-controls">
-                <a href="javascript:;" class="map-action-link" id="use-location">
-                  <span class="i18n" data-i18n="ftm-maps-search-use-location">Use my location</span>
-                  <span class="search-control-separator">|</span>
+              <div class="social-media-icons">
+                <a href="https://www.instagram.com/findthemasks/">
```
drwestco, line 215: Is this one TBD?

public/index.html
```diff
@@ -1,214 +1,237 @@
...
+                  <img src="images/instagram.svg" border="0"  />
                 </a>
-
-                <a href="javascript:;" class="map-action-link" id="reset-map">
-                  <span class="i18n" data-i18n="ftm-maps-search-reset">Reset map</span>
+              </div>
+              <div class="social-media-icons" >
+                <a class="twitter-share-button"
+                   href="https://twitter.com/intent/tweet">
```
drwestco, line 221: No target= or script action for this link?

public/index.html
```diff
@@ -1,214 +1,237 @@
...
+                  <img src="images/twitter.svg" border="0"  />
+                </a>
+              </div>
+              <!-- Reddit icon -->
+              <div class="social-media-icons">
+                <a href="https://www.reddit.com/search?q=findthemasks">
```
drwestco, line 227: No target= or script action for this link either?

public/index.html
drwestco: Can remove the target=popup attribute too, since you're handling it via onclick.

drwestco: Should have the same URL in href as you're opening via onclick, to handle middle- and right-button click actions consistently.

public/index.html
```diff
@@ -1,214 +1,237 @@
...
+      <div class="question i18n" data-i18n="ftm-add-donation-site-question">Do you work in a hospital or clinic and need PPE? Want to update or remove information?</div>
+      <p class="i18n" data-i18n="[html]ftm-add-donation-site-answer">Please <a href="https://docs.google.com/forms/d/e/1FAIpQLSfgCpK5coPVFC6rJrE7ZhimiZuDoEaL6fo6gYqxsN_FIpJZhg/viewform">fill out this form</a>. We will review it as soon as possible.</p>
+      <div class="question i18n" data-i18n="ftm-comments-questions">Got a comment or suggestion not related to data updates? Want to help?</div>
+      <p class="i18n" data-i18n="[html]ftm-contact-info">
+        Email us at contact@findthemasks.com or join us on <a target="_blank" rel="noreferrer noopener" href="https://github.com/r-pop/findthemasks">our Github</a>.
+      </p>
+      <p class="i18n" data-i18n="[html]ftm-made-with-love">made with &lt;3 in Seattle</p>
+      <a href="javascript:;" onclick="window.location='whoweare.html'+window.location.search;" class="end i18n" data-i18n="ftm-who-we-are">who we are</a>
+      <div>
+              <div class="social-media-icons">
```
drwestco, line 209: nit: class should be 'social-media-icon', singular, since you're only applying this to one icon at a time.

drwestco:
For 2, I'm not sure how quickly the new design will get going, so I wouldn't wait for it. Maybe there's room in the fixed bottom nav next to country/language selectors?

public/index.html
crayne: Of course.

public/index.html
crayne: I'll fix these -- blogspot shouldn't be there

drwestco:
I don't know why github's "ignore whitespace" review setting isn't sticky, but it makes formatting changes so much easier to deal with. Only a few changed lines in the .js file after toggling that.

public/index.html
crayne: I'm not sure what you mean.

crayne:
Re the pre-filled tweet.  According to twitter docs it would look something like this:
<
href="https://twitter.com/intent/tweet?text=Find%20out%20where%20to%20donate%20personal%20protective%20equipment">
>
How do I handle internationalization?  I see some examples in the code but they don't include urls.

nburt:
> Re the pre-filled tweet. According to twitter docs it would look something like this:
> 
> href="https://twitter.com/intent/tweet?text=Find%20out%20where%20to%20donate%20personal%20protective%20equipment"
> 
> How do I handle internationalization? I see some examples in the code but they don't include urls.

That is a great question... I don't know that you'd be able to do it with just html. The only way I can think of would be to set the href using JS and translate the text param with `$.i18n('your key')`

saikatbhadra:
Yea you would have to do it through javascript

crayne:
When I tried to use Javascript to translate the prefill text and then do an http request, I got a security error:
Access to XMLHttpRequest at 'https://twitter.com/intent/tweet' from origin 'http://local.findthemasks.com:3000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
So I have backed off the prefill idea.  I will open another issue for this.

drwestco:
You'd need to set the href attribute through script, not perform the request through XHR.

crayne:
I tried that -- I called a function in the href, then returned the url from the function, and it did not navigate to the twitter message window as it did when I included the url in the href.

crayne:
I figured it out -- I needed to change the href value in window.onload, not when the button was clicked

nburt:
@crayne almost there. Are you around to fix merge conflicts and then move the twitter href setting to a JS file (locations-list-map.js would work, you'll also need to make sure that it doesn't throw any errors on give.html where that script is also used)

crayne:
I am trying to merge the changes from this PR into the latest master, and can't figure out where to put things. There are 2 blocks of code:
1.  a window.onload script that changes the value of href in the twitter share button to point to the twitter share window with a prefilled message. I could put that at the top of the file but it would be in the body, which is weird. It will now have a call to a function in locations-list-map.js.
2.  the share buttons themselves, which were in nav-bottom, before the language and locale dropups.  Now that is all included in   {{> bottom_nav }}
Advice?

ngottlieb:
@crayne 2: that's rendering a partial that's now here: https://github.com/findthemasks/findthemasks/blob/master/views/partials/bottom_nav.handlebars

crayne:
Ah, thanks!

crayne:
All fixed -- I moved the window.onload into the js file.  Please review.

views/index.handlebars
```diff
@@ -1,109 +1,202 @@
+<!DOCTYPE html>
+<html>
```
nburt, line 2: The <html>, <head>, and <body> tags were all extracted into the main.handlebars layout and it looks like they're back in here. Since you started this before the node rewrite there have been some pretty substantial changes and I'm not really sure how to merge this. I think it would be easier to get this merged if you were to open a new PR with just the social share links. That part all looks good and for a new PR I think you'd just need to update the `bottom_nav.handlebars` partial and `locations-list-map.js` with the HTML / JS you've already written

crayne: That makes sense

crayne:
Replaced with #568

Issue #462, opened 2020-04-05
Get JavaScript errors from users 


saikatbhadra:
Install something like sentry/airbrake so we see the JavaScript errors users get.

We should hook it up to slack so developers get notified as well.

saikatbhadra:
Options:
- sentry.io
- rollbar
- airbrake

Emailed vendors to see if any of them could give us a free upgraded plan.

PR #511, opened 2020-04-08
Setup airbrake

saikatbhadra:
Sets up airbrake so all javascript uncaught exceptions are logged and sent to the #dev channel.

Closes: https://github.com/findthemasks/findthemasks/issues/462

### DIFF
--- a/public/airbrake.js
+++ b/public/airbrake.js
@@ -1,13 +1,19 @@
 // TODO we should have the environment change so we don't alert on local dev machines
+let environment;
+if (location.hostname === "localhost" || location.hostname === "127.0.0.1") {
+  environment = 'local'
+} else {
+  environment = 'production'
+}
 const airbrake = new airbrakeJs.Client({
   projectId: 266956,
   projectKey: '2c966fe44a092a4dfe2abeb19c9e112d',
-  environment: 'production'
+  environment: environment
 });
 try {
   // This will throw if the document has no head tag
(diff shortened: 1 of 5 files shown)


public/airbrake.js
```diff
@@ -0,0 +1,19 @@
+// TODO we should have the environment change so we don't alert on local dev machines
+let environment;
+if (location.hostname === "localhost" || location.hostname === "127.0.0.1") {
```
nburt, line 3: what about local.findthemasks.com?

public/give.html
```diff
@@ -9,6 +9,8 @@
   gtag('js', new Date());
   gtag('config', 'UA-161442455-1');
   </script>
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/airbrake-js/1.6.7/client.min.js"></script>
```
nburt, line 12: should we also add to stats / whoweare? Stats has some js, I don't believe whoweare does

crayne: Maybe there should be a separate channel for this.

public/airbrake.js
saikatbhadra: Done

public/give.html
saikatbhadra: We need to put all the critical js in a common js that is included in all the pages. We shouldn't have to remember to include it on each page (esp as I anticipate the new UX to give us more pages).

Issue #504, opened 2020-04-07
h2 tags should not have capitalize applied to them because different languages have different casing rules


nburt:
US title casing differs from other languages. e.g. in French you only capitalize the first word in a header.

We need to update strings to fit the country casing rules + remove the capitalize css

ngottlieb:
will change the CSS, but @nburt question re: translation -- when we change titles, will Smartling always re-translate them with first-letter capitalization as we do in the US, meaning we'll have to fix it, or will it learn the capitalization differences?

PR #512, opened 2020-04-08
Fix h2 tag capitalization

ywyw:
For subtitles, German nouns should be uppercase, English title-case, and other languages should be sentence-cased

Addresses https://github.com/findthemasks/findthemasks/issues/504


public/i18n/es-es.json
```diff
@@ -13,7 +13,7 @@
   "ftm-accepts-open-boxes" : "Se acepta cajas / bolsas abiertas",
   "ftm-ack-data-sources" : "Este sitio también incluye datos de <a href=\"https://tinyurl.com/hospitals-needing-ppe\"> https://tinyurl.com/hospitals-needing-ppe </a> , un proyecto de:",
   "ftm-add-donation-site-answer" : "<a class='add-donation-site-form'> <button style='margin-top:20px;' type='button' class='btn-sm btn-primary'> Complete este formulario </button> </a>",
-  "ftm-add-donation-site-header" : "¿Necesita EPP ( equipo de protección personal)?",
+  "ftm-add-donation-site-header" : "¿Necesita EPP (equipo de protección personal)?",
```
nburt, line 16: @ywyw did you add this in yourself or did this come in from Smartling? We should only be updating en.json and then our translation platform will update other languages. If there's an issue with es / de can you post in #languages so that the translators can update?

ngottlieb:
@nburt I had a Smartling question about this...my reading is basically, to solve this, we change the CSS...and then also manually change all the translations to the appropriate capitalization in Smartling. But seems like there's no way around having to do that any time we change something that has different capitalization protocols in different languages?

nburt:
> @nburt I had a Smartling question about this...my reading is basically, to solve this, we change the CSS...and then also manually change all the translations to the appropriate capitalization in Smartling. But seems like there's no way around having to do that any time we change something that has different capitalization protocols in different languages?

Yeah I think that's right. Process would essentially be to make the en change, and then communicate to #languages that these changes happened and that they need to be checked in other languages. Kind of a manual process with the main takeaway for me is that we shouldn't be messing with capitalization rules with css at all

ywyw:
These changes were manually made and checked via google translate.  Not sure how to use Smartling here, but I believe the edits are correct (or at least much more correct than they currently are)

nburt:
> These changes were manually made and checked via google translate. Not sure how to use Smartling here, but I believe the edits are correct (or at least much more correct than they currently are)

OK, we can follow up on the translations in the future and I'm happy to take that based off this PR. In general we shouldn't be touching non en.json files in the repo at all, sorry I haven't done a good job of documenting this process. Can you checkout those edits and we can merge and then follow up on the translations?

ywyw:
@nburt I reverted the `es.json` change

PR #513, opened 2020-04-08
add org type to map and list views; add enumerated translations for org type

### LINKED ISSUE #475: We should show the "Category" field to donors
We don't yet expose the "category" field to our donors, and we should!

The continuous records begin at row 1472, but we will be backfilling those values early this week. Let's show these to our donors!

ngottlieb:
resolves #475 

@nburt should probably look at this to confirm I took the right approach for translating this enumerated field


ngottlieb:
(and we'll also likely want to review translations once Smartling comes back with an update

nburt:
Looks great - do you think we should add some filters in the future as well? Maybe part of the next UX pass that Tina is working on

ngottlieb:
@nburt potentially... @4kbt thoughts on ability to filter by organization type? Is it needed at this point?

PR #515, opened 2020-04-08
Add line breaks between entries

saikatbhadra:
This is just a stopgap until we get designs. I dislike CSS so if anyone has better styling to recommend please feel free to add.

How it looks on mobile:

Screen Shot 2020-04-08 at 12 06 02 PM

### DIFF
--- a/public/locations-list-map.js
+++ b/public/locations-list-map.js
@@ -574,6 +574,8 @@ function getEntryEl(entry) {
         ce('p', null, ctn(translateEnumValue(entry.open_box)))
       ]);
     }
+
+    ac(entry.domElem, ce('hr', 'entry-line-break'));
   }
 
   return entry.domElem; // TODO: generate this here.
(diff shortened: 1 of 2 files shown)


drwestco:
Why? What problem are you trying to solve? Are the bold titles of each entry not sufficient?

saikatbhadra:
We have had several folks complain that the entries flow into each other. So yea there is not enough visual cues to tell where one entry ends and one entry starts.

saikatbhadra:
@ngottlieb has suggested zebra striping which is probably a better approach

drwestco:
If you want to change this to the indent method discussed in #dev, the two rules I used were:

```
.location > * {
  padding-left: 2rem;
}

/* Added to end of existing rule */
.locations-list h4 {
  padding-left: 0;
}

saikatbhadra:
I put in the indentation fix that @drwestco suggested, thanks!

PR #522, opened 2020-04-09
New geocode firebase function

susanashlock:
This function will allow us to do geocoding on Firebase and use our Maps API quota instead of our limited freebie quota that we're currently using.  Intended to be called from Apps Script using a time-based trigger.

### DIFF
--- a/functions/index.js
+++ b/functions/index.js
@@ -42,7 +42,7 @@ const COLUMNS = [
 'AA', 'AB', 'AC', 'AD', 'AE', 'AF', 'AG', 'AH', 'AI', 'AJ', 'AK', 'AL', 'AM',
 'AN', 'AO', 'AP', 'AQ', 'AR', 'AS', 'AT', 'AU', 'AV', 'AW', 'AX', 'AY', 'AZ'
 ];
-const WRITEBACK_SHEET = "'Form Responses 1'";
+const WRITEBACK_SHEET = "'Combined'";
 
 // The OAuth Callback Redirect.
 const FUNCTIONS_REDIRECT = `https://${process.env.GCLOUD_PROJECT}.firebaseapp.com/oauthcallback`;
@@ -143,7 +143,6 @@ async function annotateGeocode(data, sheet_id, client) {
           to_write_back.push({row_num, lat_lng});
           return lat_lng;
         }).catch( e => {
-          console.error("wut");
           console.error(e);
           entry[latIndex] = 'N/A';
           entry[lngIndex] = 'N/A';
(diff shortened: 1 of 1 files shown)


functions/index.js
```diff
@@ -469,6 +469,44 @@ module.exports.reloadsheetdata = functions.https.onRequest(async (req, res) => {
   res.status(200).send(html_snippets);
 });
 
+async function updateSheetWithGeocodes(country) {
+  const client = await getAuthorizedClient();
+  let data = {}
```
awong-dev, line 474: Is this used?

functions/index.js
```diff
@@ -469,6 +469,44 @@ module.exports.reloadsheetdata = functions.https.onRequest(async (req, res) => {
...
+
+  // Open relevant sheet
+  const sheets = google.sheets('v4');
+  const request = {
+    spreadsheetId: SHEETS[country],
+    range: 'Form Responses 1'
```
awong-dev, line 480: I think now it's safe to just read from Combined always?

functions/index.js
```diff
@@ -469,6 +469,44 @@ module.exports.reloadsheetdata = functions.https.onRequest(async (req, res) => {
...
+  };
+  request.auth = client;
+
+  // Transition code as we rename the output sheet from
+  // Form Responses 1 to Combined.
```
awong-dev, line 485: We should just be able to use combined in the new function and not worry about transitioning... in fact I think we've actually transitioned all sheets anyway.

functions/index.js
```diff
@@ -469,6 +469,44 @@ module.exports.reloadsheetdata = functions.https.onRequest(async (req, res) => {
...
+    response = await sheets.spreadsheets.values.get(request);
+  }
+  // Find rows that have been approved but not geocoded.
+  // Call geocoder.
+  // Fill in cells with lat, lng
+  await annotateGeocode(response.data, SHEETS[country], client);
+}
+
+module.exports.geocode = functions.https.onRequest(async (req, res) => {
+  const country = req.path.split('/',2)[1] || 'us';
```
awong-dev, line 500: Can we pull this out into a function:

get_country_from_path(req.path)

so it's not copy/pasted on 2 exports

functions/index.js
susanashlock: GB wasn't transitioned earlier tonight, but updated.  I'll verify that others are fine.

functions/index.js
susanashlock: Nope, data isn't used.  Good catch.  Removed.

functions/index.js
susanashlock: FYI, Poland appears to be in a half-baked state.  Fortunately it's not in the list of sheets in this script.

functions/index.js
susanashlock: Done.

PR #524, opened 2020-04-09
Change automatic airbrake alerting to only happen in production'

saikatbhadra:
We were creating airbrake alerts when people were testing on their local machines, this code change removes automatic alerts if the environment isn't production


ngottlieb:
👍

PR #528, opened 2020-04-10
Anonymize IPs

saikatbhadra:
This will ensure IPs sent to GA are anonymized. Following directions from here:
https://developers.google.com/analytics/devguides/collection/gtagjs/ip-anonymization


nburt:
Is there anything we need to do to scrub past data?

PR #532, opened 2020-04-10
Fill missing address and rewrite lat/lng.

awong-dev:
This makes the workflow more sane if a moderator needs to
update an address.


awong-dev:
Taggin @susanashlock as fyi.  Committing.

PR #537, opened 2020-04-11
add privacy policy pdf extract shared template

nburt:


views/privacy-policy.handlebars
```diff
@@ -0,0 +1,8 @@
+<main class="container my-5">
+  <h1 class="mb-3">Privacy Policy</h1>
+
+  <object data="https://docs.google.com/gview?embedded=true&url=https://findthemasks.com/files/ftm_privacy_policy.pdf" width="100%" height="1100px">
```
nburt, line 4: these will work on prod but file doesn't exist yet. if you sub url for the heroku preview app url it should work

PR #544, opened 2020-04-11
use partials to render country-specific donation sites

nburt:


middleware/setCurrentCountry.js
```diff
@@ -0,0 +1,12 @@
+module.exports = (req, res, next) => {
+  const originalUrl = req.originalUrl;
+  const directories = originalUrl.split('/');
+
+  if (directories.length > 2) {
+    res.locals.currentCountry = directories[1].toLowerCase();
```
ngottlieb, line 6: do we want or need to do some kind of whitelist here? (match against a list of countries that we're actually set up for?)

nburt:
This is a POC for allowing country leads to have more control over their country specific pages. Before we were using a config in JS that was getting complicated. Rather than using JS and setting the text / html content of divs on page load this approach renders different partials using country specific overrides.

Currently there is one partial (and bottom nav) that I've extracted and each partial requires its own function to pass the correct path to the view but I'm sure this can be improved upon.

For the POC, each country could override the section for what to do with large donations. Currently only Canada overrides it. I think this is simpler than our current approach but welcome to feedback and thoughts. I don't want to try and get rid of complexity by just introducing complexity somewhere else.

Other future considerations are how we handle things a site redesign and what problems could arise from partials diverging.

middleware/setCurrentCountry.js
nburt: I pretty much copy/pasted this from our FE country lookup. Do you think it's necessary to gracefully handle this? I suppose we'd just fallback to the US as the default like if there isn't a directory? At this point I think the only way this could happen is if someone manually changed the url so I'm not super worried about it

ngottlieb: fair point, doesn't do much for us at this stage

Issue #402, opened 2020-04-01
Generate filter-by items based on data instead of hard-coding them/Filter by Items do not match Google Form items - [US/CA/FR/all]


linderr:
The items we allow hospitals to check off on the Google Form do not match the Filter items. See here for US-form vs. US-filter list example. It's the same issue on Canada and France form/pages:
![image](https://user-images.githubusercontent.com/62862914/78111379-6c346480-73b1-11ea-9f60-06ffb6cebd98.png)

This mismatch does not allow us to provide full usability to this dataset. 
For example, now that the maker community is growing to produce homemade masks and 3D printed items, they need a reliable way to filter by "homemade mask" for example, which will only display hospitals that have been allowed to receive such non-medical grade items.

Although times are dire, there are institutional regulations and policies when it comes to PPE. There is a tension between frontline medical staff and hospital administration who must think about resource management as well as risk management (read: lawsuits). If we offer hospitals etc the option to select "homemade masks" on our Google Forms, we need to also make sure to help our users best navigate our data by providing that as a corresponding Filter Item.

drwestco:
Is there a way the data can tell us the list of filters to show, instead of having it hard-coded in the site script? I'm imagining a new blob of data with key/value for each item type - key would be a canonical short code used to identify the type, value would be the display text for the filter control. Each location row would also have a field with a list of keys for the items accepted at that location.

awong-dev:
@drwestco this is a good idea. The "MergeConfig" tab has this info. We need to dump and parse.

nburt:
I'd also like to integrate this with translation as well. Since we're using the MergeConfig to write enums into English we'll need to translate them into the user's selected locale when displaying them in map popups + the locations list.

cgrayson:
Suggest renaming this issue to something like, "Generate filter-by items based on data instead of hard-coding it". (The filter-by items do match the form items, for now; see my screenshot on #401.)

cgrayson:
> The "MergeConfig" tab has this info. We need to dump and parse.

I'm confused about this. When I look at the "MergeConfig" tab of the spreadsheet, I don't see much data: 

<img width="305" alt="Screen Shot 2020-04-07 at 8 12 08 AM" src="https://user-images.githubusercontent.com/5596/78673742-2cd1b080-78a8-11ea-835b-2d9c1c0bf2a0.png">

nburt:
FYI this is in PR right now: https://github.com/findthemasks/findthemasks/pull/494

[...]

Issue #401, opened 2020-04-01
Filter by Accepted Items doesn't work reliably (or at all for non-EN markets)


drwestco:
Filter by Accepted Items does a string search on the Accepting field of each listing. Each item in the filter list uses a short name to search: n95s, masks, shields, booties, goggles, gloves, sanitizer, overalls, gowns, respirators

Obviously, in a non-EN market, when the Accepting field is in the local language, this match is never going to succeed.

Even in EN, it's not reliable. See the short name 'respirators' - this will match against both N95 masks/respirators and Advanced respirators and give mixed results. 'masks' will match both N95 masks and Surgical masks. etc.

Should we just get rid of this set of filters? Is there a better solution to apply these filters?

linderr:
No, the filters are important - I was just going to create a separate Issue on them to actually add more categories to match what is being allowed through the Google Form. 

Especially now with more hospitals allowing homemade or 3D printed items, we need a way to display sites which have specify they allow for that. As much those in the maker space mean well, there are a lot of hospital policies and politics in place that prevent them, from a hospital management's perspective, to take on these items due to lack of iron-clad policies to take them on.

linderr:
The issue I wanted to raise was that currently, there is no match between Google Form Items and Filter Items. That is like allowing the user to indicate they live in New York State and having no filter tile for "NY". Here is a screen grab from US-form items vs. US-filter list. Will it a separate issue for now as I think this can provide better usability of the data:
![image](https://user-images.githubusercontent.com/62862914/78110780-71dd7a80-73b0-11ea-9083-857cd0bec1ef.png)

cgrayson:
From what I can tell, the problems described in this issue have already been fixed: the "filter by accepted items" values are more unique than what @drwestco described, and the values have been translated beyond EN. 

The issue raised by @linderr also seems resolved, as the filter-by items now match the request form: 

![issue401](https://user-images.githubusercontent.com/5596/78671305-bbdcc980-78a4-11ea-872e-76b638701912.jpg)

PR #546, opened 2020-04-11
rework filtering to dynamically generate filters based on data and wh…

### LINKED ISSUE #493: Filters should be dynamic by country
Different countries have different accepted ppe in their forms. We need to make the filters that appear below the map dynamic so that, for example, walkie talkies don't appear in France where they are not in the form.

ngottlieb:
…itelist in formEnumLookups

I believe this resolves #493, #402, and #401 (although 401 I *think* was already solved)


public/formEnumLookups.js
```diff
@@ -1,79 +1,66 @@
...
-    isSet: false
+    name: 'ftm-accept-item-face-shields'
+  },
+  'face shields - 3d-printed/makerspace': {
+    name: 'ftm-accept-item-face-shields-makerspace'
   },
-  'faceshields - 3d-printed or makerspace (specify type in other)': {
-    name: 'ftm-accept-item-face-shields-makerspace',
-    isSet: false
+  'face shields - 3d-printed or makerspace (specify type in other)': {
```
nburt, line 23: Do you think that the stuff in parens is actually necessary here? I think it makes sense on the google form but for a filter I think it's weird.

Not a blocker and could just be a follow up

ngottlieb: I agree, but that's how it's coming in from the forms -- in the US we have the parenthetical and in Canada we don't. Kind of annoying but I was told that it'd be more work to back-fit the data in the sheets than just to double up here.

I could change the translation so that they look the same though in `en.json` -- since `(specify type in other)` makes no sense in a filter

nburt: yeah I think that updating the translation would be great

ngottlieb: @nburt I reworked so that they just use the same translation as the one *without* "(specify type in other")

Issue #561, opened 2020-04-12
Add link to Mask Match in "How to Donate PPE" 


cchang-adi:
Mask Match has reached out to form a partnership, and we've agreed to mutually link each other's sites.

They would like to be included in the "How to Donate PPE" section on our landing page. Suggested text:

**Want to mail your masks?** 
Our partners at https://www.mask-match.com/ can help.
 

BTW - if we're editing the front page, there's a missing link at the very bottom of this section, at 

**Have you already donated?**
Tell us how you helped by posting on .

linderr:
Also putting together a Collaborators page where we could list them again! :)

saikatbhadra:
Is mask match international?  If not please list the countries to show for
this link.
-- 
-Saikat
saikat.bhadra@gmail.com
(832) 654 1735
https://www.linkedin.com/in/saikatbhadra

linderr:
Pretty sure it’s only US. @cchang-adi ?

cchang-adi:
From their FAQ page, it seems they’re mainly US and Canada, with some partners in Brazil and Finland. https://www.mask-match.com/faq

linderr:
That's awesome! We need a Brazil and Finnish team :)

saikatbhadra:
This is closed, thank you @ywyw

linderr:
hahah nooo.

PR #562, opened 2020-04-12
Add link for mask match

ywyw:
Addresses https://github.com/findthemasks/findthemasks/issues/561

Will need Smartling translations for other languages.

### DIFF
--- a/views/index.handlebars
+++ b/views/index.handlebars
@@ -90,6 +90,10 @@
           </li>
         </ol>
 
+        <div class="question i18n" data-i18n="ftm-want-to-mail-question">Want to mail your masks?</div>
+        <p>Our partners at <a target='_blank' rel='noreferrer noopener'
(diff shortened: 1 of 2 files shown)


views/index.handlebars
```diff
@@ -90,6 +90,10 @@
           </li>
         </ol>
 
+        <div class="question i18n" data-i18n="ftm-want-to-mail-question">Want to mail your masks?</div>
+        <p>Our partners at <a target='_blank' rel='noreferrer noopener'
```
nburt, line 94: this 2nd phrase is missing i18n. You can insert as html including the `a` tag by prefix the key with `[html]` in the data attribute. e.g. `data-i18n="[html]ftm-my-key-name"`

nburt:
@ywyw can you open a PR from the main repo rather than a fork? For some reason it seems like Smartling doesn't pull in changes to en.json when the PR is a fork

ywyw:
Gotcha, I'll close this and reopen another PR from master

PR #563, opened 2020-04-13
Test if cutting out content will improve conversion %

saikatbhadra:
I would like to bring up the map higher into the page to test if this improves conversion rates so I have cut out some content. Here are the changes: 
- removes the imagery of PPE (I'd like to put this on another page eventually)
- remove the tag line and makes the CTA a direct explanation of what our site does.
- remove line to share with industry friends since we will have share links
- removes the facebook share line since we have a PR to handle share links

The net effect is this brings our two main interactions i) adding a donations site ii) browsing the map, much further up in the page. On desktop you can reac the map in about 2 pages of scrolling now. I'd like to test out if this change will result in a net increase in the % of people who view a donation site on our page. If doesn't improve things I could revert the changes.

I only cut content and didn't edit any phrases to make them more concise since this would cause more work for the translation teams. I also didn't do any CSS stuff since I suck at that, but happy to take suggestions there.

Almost all of these changes are a part of the future design proposed that we will get in the future.

### DIFF
--- a/views/index.handlebars
+++ b/views/index.handlebars
@@ -11,40 +11,10 @@
           <!-- <h1>#findthemasks</h1> -->
           <div class="member-of i18n" data-i18n="[html]ftm-index-site-coalition">A member of the <a target="_blank" rel="noreferrer noopener" href="https://www.ppecoalition.com/">PPECoalition</a></div>
       </div>
-
-      <div class="subhead">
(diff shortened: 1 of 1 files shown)


saikatbhadra:
image
image

saikatbhadra:
Current conversion rates based on direct traffic (no give.html):

Screen Shot 2020-04-13 at 12 07 55 PM

saikatbhadra:
Around 23-26% on weekdays

PR #568, opened 2020-04-14
New version of share_links PR

crayne:
To fix merge messiness


public/locations-list-map.js
```diff
@@ -1090,11 +1090,11 @@ function updateStats($elem, count, states) {
  *
  * Result:  "1.000,2"
  */
-function number_format(number, decimal_places, dec_seperator, thou_seperator) {
+function number_format(number, decimal_places, dec_separator, thou_separator) {
```
ngottlieb, line 1093: is it just me or is there no diff here? Any idea why Github is being so weird?

public/locations-list-map.js
```diff
@@ -1112,16 +1112,21 @@ function number_format(number, decimal_places, dec_seperator, thou_seperator) {
...
     let j = h.substr(0, (h.length % 3 == 0) ? 3 : (h.length % 3));
     f[0] = j + f[0];
   }
-  dec_seperator = (decimal_places <= 0) ? '' : dec_seperator;
-  return f[0] + dec_seperator + f[1];
+  dec_separator = (decimal_places <= 0) ? '' : dec_separator;
+  return f[0] + dec_separator + f[1];
+}
+
+window.onload = function(){
```
ngottlieb, line 1129: Apologies if this was talked about in your original PR but -- I think this code can go here instead:

https://github.com/findthemasks/findthemasks/blob/master/public/locations-list-map.js#L374-L389

that block of code is already within a `$(document).ready` call and, within that, inside an `i18n:ready` event, so it would ensure that `i18n` is fully loaded by the time it tries to execute. It's possible your code *could* fail (though obviously it's working now when I'm testing it).

ngottlieb:
Oh, last thing, I'm not seeing a Smartling translation PR for this PR... @nburt do you think this is because Smartling decided it had already done this work for the other PR -- even though the other PR hasn't been merged?

public/locations-list-map.js
crayne: Github is fine.  I corrected the spelling

ngottlieb: ohhhh. Good catch! Somehow I missed that looking at it.

public/locations-list-map.js
crayne: I didn't notice that -- I'll change it.

PR #579, opened 2020-04-14
add proxy for csv file

### LINKED ISSUE #576: data-us.csv is gone
In the past, we surfaced a CSV export of our US dataset at data-us.csv  (or perhaps data.csv). This morning, one of our partners reported it missing.

Can we bring it back? (Also, org_type was missing from CSV in the past, can we make the CSV columns match the json columns?)

ngottlieb:
resolves #576

I wasn't sure what the best route to go on this was. I don't think this endpoint gets hit a lot. This turned out to be really simple from a code perspective, but it adds a dependency and may have other issues I haven't thought of. It seems to work though (and didn't break anything else).


awong-dev:
Seems good enough. I hand-rolled the middleware last time but this works.

PR #580, opened 2020-04-15
WIP : Top navigation bar for index v2

drwestco:
First pass at creating a top nav bar to replace existing bottom nav + header. Could use additional UX design + dev work.

image


public/i18n/en.json
```diff
@@ -168,5 +168,9 @@
   "ftm-thanks-smartling": "Smartling for access to their Translation Managment Platform",
   "ftm-thanks-google": "Google for Maps API credits",
   "ftm-thanks-slack": "Slack for a Pro account",
-  "ftm-tweet-share-button": "Find the closest place to donate masks and other PPE at www.findthemasks.com"
+  "ftm-tweet-share-button": "Find the closest place to donate masks and other PPE at www.findthemasks.com",
+  "ftm-nav-about": "About",
+  "ftm-nav-partners": "Partners",
+  "ftm-nav-volunteer": "Volunteer",
+  "ftm-nav-addplace": "Add a place"
```
nburt, line 175: Should this be "Site"? I think that's the language we currently use

nburt:
I think the dropdowns are a little crowded on mobile (below, maybe some margin below the logo and padding for the selected item), minor comment on copy, overall I think this looks great.

image

I'm a little worried about social getting pushed out of sight. I'm not sure how many people are clicking on them but this would remove them from view. What do you think about leaving in a bottom nav or putting somewhere above the map until designs for that are ready?

linderr:
CC @tinabeans @sumitsekhri

linderr:
@nburt Sumit posted some mock ups on #ux-design for mobile and it's fairly different than the desktop version! So maybe this is not an issue. @sumitsekhri ?

tinabeans:
on mobile we could just show the flag and omit the country name from the dropdown trigger, and abbreviate the language to its ISO code: https://www.loc.gov/standards/iso639-2/php/code_list.php - that way they could sit on the same line?

drwestco:
Fixed the dropdown padding for mobile.

public/i18n/en.json
linderr: @drwestco we settled with "I need PPE" CC: @tinabeans

drwestco:
Resolved the latest merge conflicts from the translations, so button now shows the expected "I need PPE" text. What else needs to be done before merging this to master? 

Need to get the social icons up above the pit of endless scrolling? I really don't want to reintroduce fixed bottom navigation bar just for this (or at all - letterboxing = bad!). I can whack them in just above the HR before the Donation Sites header, though the text above it ("got a comment" through "who we are" probably should come with them in that case). Other suggestions?

tinabeans:
It looks like very few people click on the social sharing buttons. This could either be because they are hard to find, or because they don't need them. We could put them in the header under a dropdown that says "Share," like so:

Header Drop Down

drwestco:
Maybe we can integrate the share links with the not-yet-implemented About feature?

linderr:
@crayne also just implemented them so that could be another explanation of why they are ill-used

nburt:
My vote is to do what @drwestco recommended and then just remove the footer so we can get the navbar out. I think that what we have is a step forward and the map still isn't above the fold so adding another section isn't too bad from that perspective.

From there I think we should integrate it into the v2 design. I wonder if the sharing could be a CTA related to sites? Something along the lines of "Share your hometown hospital to amplify their donations" (with the caveat that we'd need to figure out how to share markers / locations and have them load highlighted based off the link)

linderr:
Love that idea @nburt I had asked for a similar logic previously to provide regional-specific default zooms for our PPE runner collaborators who operate in a select radius who embed our map attempting to show that. 

For now I'm just giving them ?state=xx URLs which is not ideal but narrows it down from a national, default view. @graemer had provided me with a logical outline of: https://findthemasks.slack.com/archives/C010L1HQ4TC/p1585932714187700?thread_ts=1585930078.181900 

Now with v2 I am wondering whether we can just output the lat/lng columns per site as a "Share this Hospital" CTA as you suggested and centre the default view on the lat/lng selected/shared?

drwestco:
Removed the separate footer section. Moved the unique content (who we are + social icons) up above the map.

linderr:
so exciting. Thanks @drwestco !! There is double up on the CA site for the footer section. Will make a new issue but thank you for this update and part one of the v2 roll out !!!

PR #582, opened 2020-04-15
add i18n to node + translate social sharing tags

nburt:


middleware/setBananaI18n.js
```diff
@@ -0,0 +1,61 @@
+const url = require('url');
```
nburt, line 1: this is currently pretty much a copy/paste from the FE i18n.js file in terms of locale detection + the map of locales / files, just exported as node middleware instead.

Issue #577, opened 2020-04-14
404 redirect is almost usable


4kbt:
Visiting a malformed FTM URL, like https://findthemasks.com/godevsgo.html 

redirects to https://findthemasks.com/godevsgo.html/

Which looks like this (note missing flag image): 
![image](https://user-images.githubusercontent.com/3980568/79243246-39b34e80-7e2a-11ea-94fd-ec924cbf3fd6.png)

And attempts but fails to load donation pages like this: 

![image](https://user-images.githubusercontent.com/3980568/79243312-53ed2c80-7e2a-11ea-9c04-a97c7d56b58e.png)

A redirect to a functioning site like the US, or to a global directory of FTM pages, might help to get the reader where they need to go?

nburt:
pretty sure the problem is this line in server.js: `app.use('/:countryCode', router);`

Essentially this is matching as a "country code" and trying to grab the svg / markers for `godevsgo.html`.

PR #585, opened 2020-04-15
redirect to root on 404s

nburt:
Fixes https://github.com/findthemasks/findthemasks/issues/577


awong-dev:
Thank you for fixing! This was getting to me.

PR #604, opened 2020-04-16
Feature/map nav query params

### LINKED ISSUE #603: Add map navigation query params that allow sharing links to places
Adding:

- `q`: a map search query
- `coords`,`zoom`: obvious enough

There was some discussion about whether `lat,lng,zoom` could just be passed as the `q` param. I'm not so sure about doing this. If you search `lat,lng` in our existing search, it works. But I'd like `q` to leverage our existing search mechanism, while specifying `lat,lng,zoom` should go to exactly where is specified, regardless of whether or not 3 markers are included.

This sets the stage for allowing users to share a link directly to what they're looking at either by:

- having a "share this map" button that generates a URL with query params in it
- auto-updating browser location using the map position when users interact with the map

This accomplishes one of the feature requests embedded #601 though there may be further steps

ngottlieb:
resolves #603

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ Join the slack! <https://join.slack.com/t/findthemasks/shared_invite/zt-czdjjznp
 <iframe style="width: 100%; height: 800px; border: none;" src="https://findthemasks.com/give.html"></iframe>
 ```
 
-- We also support state specific data views, hiding the map, hiding the filters through query params, and specifying locale:
+- We also support state specific data views, hiding the map, hiding the filters through query params, initializing on specific locations, and specifying locale:
 
 ```html
 ?state={CA/WA/NY/etc}
@@ -42,6 +42,8 @@ Join the slack! <https://join.slack.com/t/findthemasks/shared_invite/zt-czdjjznp
 ?hide-list={true/false} (also hides filters)
 ?hide-search={true/false} (beta)
 ?locale={see available locales in dropdown at https://findthemasks.com}
+?q={map search term: '100 Fake Road, Kansas, MO'}
+?coords={lat,lng}&zoom=12 // zoom only works in concert with coords
 ```
 
 All boolean parameters default to false (unless they're in beta).
--- a/public/locations-list-map.js
+++ b/public/locations-list-map.js
@@ -684,7 +684,7 @@ function initMapSearch(data, filters) {
   const url = new URL(window.location.href);
   const searchParams = new URLSearchParams(url.search);
   const q = searchParams.get('q');
-  if (q && q.length) {
+  if (q) {
     $search.val(q);
     attemptGeocode(q);
   }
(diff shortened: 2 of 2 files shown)


ngottlieb:
Oops gonna update the README to reflect the new params as well!

public/locations-list-map.js
```diff
@@ -685,6 +680,15 @@ function initMapSearch(data, filters) {
     searchEl, { types: ['geocode'] }
   );
 
+  // initialize map search with query param `q` if it's set
+  const url = new URL(window.location.href);
+  const searchParams = new URLSearchParams(url.search);
+  const q = searchParams.get('q');
+  if (q && q.length) {
```
drwestco, line 687: Don't need to check length. Empty string is falsy.

public/locations-list-map.js
```diff
@@ -703,22 +707,9 @@ function initMapSearch(data, filters) {
...
-        // Ensure we got a valid response with an array of at least one result.
-        if (status === 'OK' && Array.isArray(results) && results.length > 0) {
-          let viewport = results[0].geometry.viewport;
-          fitMapToMarkersNearBounds(viewport);
-        } else {
-          console.warn('Geocode failed: ' + status);
-          sendEvent("map","geocode-fail", $search.val());
-        }
-      });
+      console.warn('No geometry found, attempting geocode...');
```
drwestco, line 711: Can probably get rid of this warning now.

public/locations-list-map.js
```diff
@@ -954,9 +960,31 @@ const MAP_INITIAL_VIEW = {
...
 
+function getMapInitialView() {
+  const url = new URL(window.location.href);
+  const searchParams = new URLSearchParams(url.search);
+  const coords = searchParams.get('coords');
+  // default zoom is pretty tight because if you're passing latlng
+  // you are probably trying to center on a pretty specific location
+  const zoom = parseFloat(searchParams.get('zoom')) || 11;
+  if (coords && coords.length) {
+    const latlng = coords.split(',').map(coord => parseFloat(coord));
```
drwestco, line 971: Need to do some more sanitization here. Validate you actually have 2 values, and that lat is in the range [-85..85].

README.md
```diff
@@ -42,6 +42,8 @@ Join the slack! <https://join.slack.com/t/findthemasks/shared_invite/zt-czdjjznp
 ?hide-list={true/false} (also hides filters)
 ?hide-search={true/false} (beta)
 ?locale={see available locales in dropdown at https://findthemasks.com}
+?q={map search term: '100 Fake Road, Kansas, MO'}
+?coords={lat,lng}&zoom=12 // zoom only works in concert with coords
```
linderr, line 46: @ngottlieb just curious would normal people just Google a lat/lng they want via "what's here?" and put it in there?

ngottlieb: by normal people do you mean like, you and other outreach volunteers, or a hospital administrator or something? For the former, I think yes. My quickest way to find latlng of a location is to open google maps and then just click on a spot on the map and latlng gets added both to the browser URL bar and a little label at the bottom of the page.

linderr: haha yes non-dev wizards. Okay, so same manual process - got it!

PR #605, opened 2020-04-16
Add analytics to top nav donation link

saikatbhadra:


public/locations-list-map.js
```diff
@@ -371,14 +371,18 @@ function loadOtherCountries() {
 $(function () {
   const url = new URL(window.location);
 
+  $(".add-donation-site-form").click(function() {
```
ngottlieb, line 374: Hmm, it looks like you're adding a `click` event to the same selector in two places now? Is what's going on that this top one exists in the DOM to be attached before i18n happens, and the other one doesn't get added to after i18n? Even with that, seems like line 382 would be attaching the new click event to the top-nav one as well?

ngottlieb: Maybe I'm missing something obvious though...

saikatbhadra: nope you are right, this change isn't needed. 🤦

ngottlieb: 👍

PR #607, opened 2020-04-17
Redesign V2

nburt:
Do not merge - opening to get heroku preview apps


crayne:
Since the state tiles are gone, I typed Idaho into the search box. The map focused on Idaho, but the list showed the entire US.

saikatbhadra:
Looking good! Make sure to fix the search bar area as the user scrolls through the list view.

nburt:
> Since the state tiles are gone, I typed Idaho into the search box. The map focused on Idaho, but the list showed the entire US.

@crayne I'm getting this which seems right, can you send a screenshot of what you're seeing?

image

crayne:
I am seeing what you are seeing.  But while the map centers on Idaho, the list on the left starts with Alaska and continues through all the states.  Isn't the list supposed to be filtered to show only locations in Idaho?  That's what it does in V1.

nburt:
Remaining todos:

- [ ] Implement filtering for facility type
- [ ] Filter locations list by map viewport
- [ ] sync master

public/images/icons/circle_check.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="14" height="14" viewBox="0 0 14 14" fill="none" xmlns="http://www.w3.org/2000/svg">
```
drwestco, line 1: You can encode and add these to icons.data.svg.css, to avoid separate requests for individual icons.

public/give.js
```diff
@@ -0,0 +1,1164 @@
+import toDataByLocation from './toDataByLocation.js';
```
drwestco, line 1: Does 'give' really need its own script?

views/partials/top_nav.handlebars
```diff
@@ -41,6 +41,31 @@
...
+      <li class="nav-item dropdown active">
+        <a class="nav-link dropdown-toggle i18n" href="#" id="share-dropdown" role="button"
+           data-toggle="dropdown" aria-haspopup="true" aria-expanded="false"
+           data-i18n="ftm-nav-share">
+          Share
+        </a>
+        <div class="dropdown-menu dropdown-menu-right" aria-labelledby="share-dropdown"
+             id="countries-dropdown-selector">
+          <a class="fb-xfbml-parse-ignore dropdown-item" onclick="window.open('https://www.facebook.com/sharer/sharer.php?u=http%3A%2F%2Fwww.findthemasks.com%2F&amp;src=sdkpreparse','popup', 'width=600,height=800');return false;" href="https://www.facebook.com/sharer/sharer.php?u=http%3A%2F%2Fwww.findthemasks.com%2F&amp;src=sdkpreparse">
+            <div class="icon icon-facebook"></div> Facebook
```
drwestco, line 53: The background of the facebook icon disappears, since we picked the same shade of blue for dropdowns + secondary header. We should probably pick a different color.

public/locations-list-map.js
```diff
@@ -476,19 +523,6 @@ function renderNextListPage() {
   const children = [];
 
   locationsListEntries.slice(renderLocation, renderLocation + 40).forEach(function (entry) {
-    // Add city/state headers
```
drwestco, line 479: Are we sure we want to get rid of city/state headers? It's weird for the list to jump from AK to AL with no break in between. May be a non-issue in the future once list respects map view, though I still think it will be weird when zoomed out to view country-wide or even multi-city results.

public/locations-list-map.js
```diff
@@ -507,17 +541,18 @@ function googleMapsUri(address) {
 
 function getEntryEl(entry) {
   if (!entry.domElem) {
-    entry.domElem = ce('div', 'location');
-    ac(entry.domElem, ce('h4', null, ctn(entry.name)));
+    entry.domElem = ce('div', 'location py-3');
+    const header = ce('div', 'd-flex');
+    const headerHospitalInfo = ce('div', 'flex-grow-1');
+    const headerOrgType = ce('div', 'flex-grow-1 d-flex justify-content-end text-pink');
```
drwestco, line 547: Hopefully there's a better way to do the layout than hard-coding these specific classes?

public/locations-list-map.js
```diff
@@ -507,17 +541,18 @@ function googleMapsUri(address) {
 
 function getEntryEl(entry) {
   if (!entry.domElem) {
-    entry.domElem = ce('div', 'location');
-    ac(entry.domElem, ce('h4', null, ctn(entry.name)));
+    entry.domElem = ce('div', 'location py-3');
+    const header = ce('div', 'd-flex');
+    const headerHospitalInfo = ce('div', 'flex-grow-1');
```
ngottlieb, line 546: I mentioned this in Slack but -- maybe this is an opportunity to replace this logic with handlebars templating?

public/locations-list-map.js
linderr: Agreed! @tinabeans

nburt: I'm thinking that once the list respects the map view it should be good enough for launch but ya I agree that city/state headers could be nice

public/give.js
nburt: I moved it out so that I could focus on index and stuff was starting to break as the filters changed. The one thing I'm worried about with 'give' is that it's already embedded in a bunch of other sites so I want to make sure we don't break anything by adjusting the height of the page too much so I didn't wanna change the filters in the first pass.

public/locations-list-map.js
nburt: I like Nick's idea of handlebars templates rather than building it in js like this. I think I'd also lean towards refactoring to that after v2 is out, thoughts?

views/partials/top_nav.handlebars
nburt: updated to a lighter blue version of the fb icon

public/give.js
nburt: @drwestco took a stab at consolidating back here: https://github.com/findthemasks/findthemasks/pull/624

We lose out on state filters but with the list synced with the map I think that's OK?

ngottlieb:
@nburt the organization type filter displays even when there are no org types (France) and is a bit weird, esp. since there's no label. Can we exclude when there are no options?

nburt:
> @nburt the organization type filter displays even when there are no org types (France) and is a bit weird, esp. since there's no label. Can we exclude when there are no options?

updated this to remove the dropdown. I didn't do it for accepted items cause I think that'll always have that

saikatbhadra:
Putting remaining todos for FAQS 

- [ ] Try to recycle the i18n logic and content for the 5 questions above (instead of recreating the content, re-triggering smartling, and retranslating)
- [ ] Fix the link to the FAQ on the index.html and give.html pages
- [ ] Setup the html for i18n (add classes and update `en.json`
- [ ] Give the page a little CSS love so there is better line spacing in each of the areas

crayne:
I did a search for Chicago.  The map was correct but the first two listings were:

image

drwestco:
Desert Care + Allergy Sleep + Encompass Health have bad or missing values in the lat/lng fields. Not sure why they get bumped up only when the map view zooms in close, though.

Also noticed the list doesn't update when map view recenters after clicking on a marker, but only when I directly pan/zoom the map.

PR #608, opened 2020-04-17
Localize contact info and made with love

maplebird:
Localize contact info so we can set a custom email address if required and add a custom "Made with love" note.

Currently just for Canada.

Request from @linderr


nburt:
@maplebird @linderr fyi pretty sure the heart is broken for all foreign languages
image

I'll go into smartling and fix, translations will update next time someone makes a change to en.json

nburt:
wait jk! not broken just a space between the < and 3 which is probably ok for a little bit

linderr:
When they’re together it becomes a heart emoji! Should be HTML URL encoding reference standard? Sorry we were messing around last night trying to learn and probably put a space by accident

nburt:
Not your fault, it's google translate's :) I don't think this is a huge deal but we can have translators review and fix I suppose?

Issue #606, opened 2020-04-17
Embedded map-link should open a new tab.


4kbt:
From the FTM-PPE cross-slack channel:

Zach Peery-Wolf: "was doing some front end testing and when I click on the "Made in collaboration with #findthemasks. Want to join the #findthemasks map? Need to update or remove info? Complete This Form" button it's opening into a frame where I can't scroll down to the bottom of the form to complete it. Can you get it to open a new tab?"

This would also resolve @linderr 's concern:

"Interesting, and there is no way to return to the embed map after someone hits "complete this form" https://theppedrive.com/ppe-drive-listing "

Including @awong-dev .

4kbt:
Relevant URL: https://getusppe.org/give/

This is working for me on my desktop screen, but apparently not for all users.

4kbt:
Hm. Working for me on mobile, too.   Will ask for clarification.

4kbt:
Error apparently happened on this config: "Firefox 75.0 (64-bit) Windows 10 Home. But I just cleared my cookies and that fixed it."

We should probably still go for the new-tab, just to be safe?

linderr:
https://github.com/findthemasks/findthemasks/pull/571

Misses target="_blank" in ftm-add-donation-site-answer but exists in all other xx.json

saikatbhadra:
closed as from the discussion on the PR this is no longer necessary

PR #612, opened 2020-04-18
Change the embed page's add/update donation site button open a new tab

saikatbhadra:
Closes https://github.com/findthemasks/findthemasks/issues/606

### DIFF
--- a/public/i18n/en.json
+++ b/public/i18n/en.json
@@ -12,7 +12,7 @@
   "ftm-accepting": "Accepting",
   "ftm-accepts-open-boxes": "Accepts Opened Boxes/bags",
   "ftm-ack-data-sources": "This site also includes data from <a href=\"https://tinyurl.com/hospitals-needing-ppe\">https://tinyurl.com/hospitals-needing-ppe</a>, a project by:",
-  "ftm-add-donation-site-answer": "<a class='add-donation-site-form'><button style='margin-top:20px;' type='button' class='btn-sm btn-primary'>Complete this form</button></a>",
+  "ftm-add-donation-site-answer": "<a class='add-donation-site-form' target='_blank'><button style='margin-top:20px;' type='button' class='btn-sm btn-primary'>Complete this form</button></a>",
   "ftm-add-donation-site-header": "Do You Need PPE?",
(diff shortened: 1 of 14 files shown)


public/i18n/en.json
```diff
@@ -12,7 +12,7 @@
   "ftm-accepting": "Accepting",
   "ftm-accepts-open-boxes": "Accepts Opened Boxes/bags",
   "ftm-ack-data-sources": "This site also includes data from <a href=\"https://tinyurl.com/hospitals-needing-ppe\">https://tinyurl.com/hospitals-needing-ppe</a>, a project by:",
-  "ftm-add-donation-site-answer": "<a class='add-donation-site-form'><button style='margin-top:20px;' type='button' class='btn-sm btn-primary'>Complete this form</button></a>",
+  "ftm-add-donation-site-answer": "<a class='add-donation-site-form' target='_blank'><button style='margin-top:20px;' type='button' class='btn-sm btn-primary'>Complete this form</button></a>",
```
nburt, line 15: can you also add `rel="noopener noreferrer"` since you're adding `target="_blank"`?

https://developers.google.com/web/tools/lighthouse/audits/noopener

jamesarosen: Is it possible to move the `<a>` and `<button>` into the `.handlebars` file? That would make this easier to maintain and less prone to inconsistencies across locales.

nburt: ya that would be better, I think there used to be text before the `<a>` that was removed recently.

ngottlieb: @saikatbhadra ping ^^

saikatbhadra: sorry i forgot about this, how do I do variable substitution with i18n

saikatbhadra:
hey @4kbt  I forgot why we needed this, do we actually need the form link in the embed page to open a new tab? why can't it redirect on the same page?

4kbt:
I believe that the acute need has now been resolved.   The problem was that
GUPPE's embed wasn't tall enough  for people to fill in the Google Form
from within the GUPPE page.  They could fill it in, but not reach
"Submit".   Forcing a new tab was a shotgun approach to solving the problem.

4kbt:
The GUPPE embed works okay right now on my desktop and mobile.

saikatbhadra:
closing for now then

Issue #602, opened 2020-04-16
Create a Volunteer/How Can I Help? page


4kbt:
I'm imagining something like findthemasks.com/volunteer 

We keep getting requests for people to help us out. Having a central location for us to advertise timely needs and for prospective volunteers to visit might help us to keep it organized, both internally and externally.

Strawman sections:

Thank you!: Thank you for your interest in helping FindTheMasks, etc.

Needs: enumerates current needs for volunteers

Contact: Provide a simple contact form for prospective volunteers to register their interests, time-availability, and skills.

linderr:
I agree! Stefanie, Jane and I are working on the About, FAQs, Partner and Join Us pages for text/content. I noticed that the "Join Us" real estate recently got replaced by the Share dropdown in v2.. @tinabeans where shall we put the volutneer page @4kbt is suggesting now?
https://findthemasks.slack.com/archives/C010J52HXPD/p1586910401413400?thread_ts=1586462363.328900

saikatbhadra:
I've setup a skeleton page. I filled out the dev section, can I get some other content?

linderr:
@saikatbhadra can you DM me your gmail on Slack? We are working on content for all the new subpages but waiting for new UI/UX dev HeatherQ to mock up designs this weekend

PR #614, opened 2020-04-18
Setup volunteer page

saikatbhadra:
Screen Shot 2020-04-17 at 10 42 04 PM

Sets up skeleton for volunteer page. 

Closes: https://github.com/findthemasks/findthemasks/issues/602

### DIFF
--- a/server.js
+++ b/server.js
@@ -54,7 +54,7 @@ router.get(['/', '/index.html'], (req, res) => {
   });
 });
 
-router.get(['/', '/volunteer.html'], (req, res) => {
+router.get('/volunteer', (req, res) => {
   res.render('volunteer', {
     ogTitle: res.locals.banana.i18n('ftm-index-og-title'),
     ogUrl: `http://${req.hostname}${req.originalUrl}`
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -54,6 +54,13 @@ router.get(['/', '/index.html'], (req, res) => {
   });
 });
 
+router.get(['/', '/volunteer.html'], (req, res) => {
```
nburt, line 57: why the '/' here? shouldn't that be '/volunteer'? I think we can also get rid of the '/volunteer.html' bit and the array match. That was mostly for backwards compatibility for index.html and give.html in case they were linked/embedded from other sites

saikatbhadra: fixed

server.js
```diff
@@ -54,6 +54,13 @@ router.get(['/', '/index.html'], (req, res) => {
   });
 });
 
+router.get('/volunteer', (req, res) => {
+  res.render('volunteer', {
+    ogTitle: res.locals.banana.i18n('ftm-index-og-title'),
+    ogUrl: `http://${req.hostname}${req.originalUrl}`
```
nburt, line 60: can you add ogDescription too?

jamesarosen: Maybe also change `http` to `${req.protocol}` so it's `https` in production and `http` in dev.

Issue #610, opened 2020-04-17
About subpage does not have v2 header 


linderr:
About subpage is now available on the new v2 header that @drwestco coded. However the subpage does not have the header.

You get to this subpage and you're trapped. No header to go back to the main page/map.

linderr:
![image](https://user-images.githubusercontent.com/62862914/79598981-a3855f80-8099-11ea-8cbe-68b76552bc05.png)

![image](https://user-images.githubusercontent.com/62862914/79599066-bdbf3d80-8099-11ea-9446-8616983c55a0.png)

We have more subpages coming (About, FAQs, Partnerships, Join Us). All subpages must have the v2 header :)

linderr:
@ywyw thank you!

PR #622, opened 2020-04-19
Add header nav to whoweare, fixup i18n issues

ywyw:
Addresses https://github.com/findthemasks/findthemasks/issues/610

1. Refactored top nav rendering to its own module
2. i18n is actually not functioning on the "Who we are" page currently due to some outdated paths, fixed that up
3. Added navigation back to home page from "Who we are" by making the FTM span clickable.


nburt:
@ywyw thanks for splitting these files apart! This will be super helpful. I just fixed a few errors and will merge

linderr:
Thank you @ywyw !! Always great work from you --- ah. Nate just merged. Yay!
We have several new subpages which will all require this logic! Thank you for separating and fixing

PR #624, opened 2020-04-19
Consolidate give js

nburt:


nburt:
This is how it looks at https://getusppe.org/give/
image

Issue #636, opened 2020-04-20
Data blog launch checklist


4kbt:
- [x] Surface HTML for comments
- [x] Get sign-off from #comms on content
- [x] Get sign-off from #comms on author list
- [x] Solicit help from #dev on best way to bring it in as a PR. 
- [x] Get #ux  / #dev assistance on html style-sheets
- [x] Figure out how it might fit into V2 styling
- [x] Are Figures 3 and 4 degenerate? (@4kbt thinks not, but smarter people do)
- [x] Integrate Lee's changes
- [x] Integrate Detwiler's changes
- [x] Proofread / styling when integrated into site, but unlinked
- [x] Create permalink (blog/2021-04-21-Data-Insights)
- [x] Make facebook share-link reference permalink
- [x] Have a dev see if there's a way to make social-media links flow in through handlebars (optional)
- [x] Cross fingers that analysis is correct-enough.
- [ ] Release
- [ ] Go public, drive publicity.

4kbt:
Document surfaced here for comment: https://faculty.washington.edu/cah49/ftmResearchNote.html

4kbt:
Closing -- 'cause it's done!

PR #639, opened 2020-04-21
Mergeable with caveats: Data-blog article #1. See #636 for checklist.

4kbt:
Creating PR at @awong-dev's suggestion in order to see how things roll in Heroku.


views/datablog.handlebars
```diff
@@ -0,0 +1,330 @@
+<div class="translated-content">
+  {{> top_nav }}
+
+  <div class="container mt-5">
+    <div class="row">
+      <div class="col">
+	<head><title>Insights from FindTheMasks-US Data</title>
```
nburt, line 7: happy to update this if needed but you'll need to set this in the layout. It'll need to be passed through as a local (like "ogTitle") and then the static.handlebars file will need to use it or fallback to a default title.

nburt:
This looks awesome! Given that this is a US focused analysis I'm guessing we don't need to translate it right?

4kbt:
Correct; I haven't begun to think about translating it.

views/datablog.handlebars
4kbt: You're going to be able to do this sort of thing much more efficiently than I can. If we make a second article, I'll be able to be a copy-paste monkey the next go-round.

4kbt:
PR is now mergeable with caveats. See #636 for the remaining tasks.

PR #640, opened 2020-04-21
Add in webpack to build pipeline. Enable hot reload.

### LINKED ISSUE #166: Support older browsers

awong-dev:
With this, webpack is integrated into the heroku build pipeline for both
CSS and Javascript. This allows transpiling the JS down into something
more compatible with older browsers like IE.

As a bonus, the webpack-dev-middleware has been added into the express
server for development mode with hot reload enabled. This means with
"npm run dev", developers will get 2 features:

  (a) the webpage with auto-reload when client JS is changed
  (b) style sheet changes are deployed without reload.
  (c) syntax errors in JS or css are displayed in an overlay in the
      browser.

It's possible to configure it so that the page does not reload when the
JS is changed, but given that there is no tie-in to rerender the
map list on reload, it doesn't help.

Note also that the css file is extracted multiple times. It's not a big
deal for now. Just always use generated/style.css.

fixes #166


awong-dev:
@drwestco oh, I missed it was dev config only. That's expected actually.

I'm using an inline sourcemap in dev cause for iterative development, the sourcemap creation is faster.  So the .js file has the full map in an eval blob which is HUGE.

For production, here's the output of my generated directory:

```
3038 i18n.entry.js
11353 i18n.entry.js.map
5870 locations-list-map.css
8003 locations-list-map.css.map
20966 locations-list-map.entry.js
76328 locations-list-map.entry.js.map
5562 render-nav.entry.js
17082 render-nav.entry.js.map
5857 style.css
7990 style.css.map
1019 style.entry.js
4643 style.entry.js.map
```

i18n.entry.js went up by .6kb over i18n.js.  However, localtion-list-map-entry.js went down 16kb compared to location-list.map.js.

I bet this is just due to some inlining that's occurring. I don't think i18n.entry should exist as an entrypoint. In fact, I wonder if client-side i18n even makes sense anymore.  @nburt

Issue #635, opened 2020-04-20
V2 hangs loading location sites in Chrome


crayne:
This does not happen in Safari.  It does not happen in Firefox.  In Chrome, it does not happen in a dev environment.

nburt:
Does this happen every time for you? I don't get it in Chrome v 80.0.3987.163. Is there anything in the console?

crayne:
It happens every time for me.  I am running Chrome v 81.0.4044.113.  In the console:
Uncaught SyntaxError: The requested module './getCountry.js' does not provide an export named 'default'

crayne:
I think this line in getCountry.js

```
export function getCountry() {
```
should be 

```
export default function getCountry() {
```

nburt:
hmm yeah I'm seeing that in our Airbrake logs. I wonder if this could be a caching issue? I can't seem reproduce it. I can only find named imports of getCountry in the codebase:

![image](https://user-images.githubusercontent.com/5788989/79868873-92e12c00-83ae-11ea-8e01-94155738ff50.png)

![image](https://user-images.githubusercontent.com/5788989/79868887-98d70d00-83ae-11ea-8e3b-5f8f68529754.png)

Can you let me know if you're still seeing this issue?

crayne:
Unfortunately, yes.  I did clear the cache.  I'm mystified about the difference in behavior between dev and production.

saikatbhadra:
Link to the airbrake issue: https://find-the-masks.airbrake.io/projects/266956/groups/2726861397683913057?tab=overview

saikatbhadra:
@nburt I believe this is the issue: 

https://stackoverflow.com/questions/54590951/uncaught-syntaxerror-the-requested-module-add-js-does-not-provide-an-export

crayne:
Actually I looked at the source for locations-list-map.js.  In the dev console  of findthemasks.com the import for getCountry is:

import getCountry from './getCountry.js' 

and the import for getCountry in the source of my pull from master is: 

import { getCountry } from './getCountry.js'  

so that would imply a caching problem, no?

To confirm this, if I run the dev version and look at the console, the import for getCountry in the dev console is:

 import { getCountry } from './getCountry.js';

My dev version runs correctly.

crayne:
www.findthemasks.com.har

PR #643, opened 2020-04-21
add query string versioning where possible to mitigate caching issue

ngottlieb:
resolves https://github.com/findthemasks/findthemasks/issues/635 (at least, first solution for it, webpack versioning will be better)

### DIFF
--- a/server.js
+++ b/server.js
@@ -9,6 +9,7 @@ const formatFbLocale = require('./utils/formatFbLocale');
 require('dotenv').config();
 const https = require('https');
 const { createProxyMiddleware } = require('http-proxy-middleware');
+const herokuVersion = process.env.HEROKU_RELEASE_VERSION;
 
 const app = new express();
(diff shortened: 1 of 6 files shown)


ngottlieb:
so this works locally (i.e. `HEROKU_RELEASE_VERSION=hello npm run dev` yields the appropriate URLs), but so far, the RA's on Heroku aren't populating the env variable. I think this is because the labs feature @awong-dev was using hasn't been enabled on review apps...but unfortunately, you [*can't* enable labs features with app.json](https://stackoverflow.com/questions/37570681/enable-heroku-labs-feature-in-heroku-review-apps). So unless we can do it in the UI somewhere, I'm not sure we can make this exact approach work on RAs.

which of course doesn't really matter given the circumstances, but I can't totally confirm this is working at this point.

nburt:
worked! nice 
image

PR #646, opened 2020-04-22
Remove module attributes

drwestco:
See if removing type="module" allows older browsers in.


client/toDataByLocation.js
```diff
@@ -1,10 +1,10 @@
 export default function toDataByLocation(data) {
   const headers = data.values[1];
-  const approvedIndex = headers.findIndex( e => e === 'approved' );
-  const stateIndex = headers.findIndex( e => e === 'state' );
-  const cityIndex = headers.findIndex( e => e === 'city' );
-  const latIndex = headers.findIndex( e => e === 'lat' );
-  const lngIndex = headers.findIndex( e => e === 'lng' );
+  const approvedIndex = headers.indexOf('approved');
```
awong-dev, line 3: Shouldn't babel handle this?

drwestco: I'd think so. Still trying to decipher the babel docs regarding polyfills.

drwestco: Think I got it sorted. Will verify with good ol' IE in the morning.

awong-dev: you rock. Thank you!!!

drwestco: Turning on the polyfills via Babel does the trick. It really bloats the .entry.js files, though:

image

views/stats.handlebars
```diff
@@ -17,8 +17,8 @@
   <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery.i18n/1.0.7/jquery.i18n.language.min.js"></script>
   <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery.i18n/1.0.7/jquery.i18n.messagestore.min.js"></script>
   <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery.i18n/1.0.7/jquery.i18n.parser.min.js"></script>
-  <script type="module" src="generated/i18n.entry.js?v=v{{ version }}"></script>
-  <script type="module" src="generated/stats.entry.js?v=v{{ version }}"></script>
+  <script src="generated/i18n.entry.js?v=v{{ version }}"></script>
```
nburt, line 20: fixing this in a separate pr but public paths need a `/` at the start of the path

PR #647, opened 2020-04-22
Pins layering

sikakkar:
When there's multiple pins at a particular address, clicking on any of them shows information for all of them. I didn't remove the pins, though, because I didn't want to mess up the count / any other logic. 

image


sikakkar:
@ngottlieb resolved the merge conflicts! Thanks

PR #660, opened 2020-04-22
Downlevel compat, take 2

drwestco:
Lighter-weight approach to downlevel browser compat. Add eslint plugin to detect issues; only add specific polyfills as needed.


client/utils.js
```diff
@@ -24,3 +24,47 @@ export function ce(elementName, className, child) {
...
+      const pair = q.split('=');
+      if (pair.length > 1 && pair[0]) {
+        searchparams[pair[0]] = pair[1] || '';
+      }
+    }
+
+    this.protocol = parser.protocol;
+    this.host = parser.host;
+    //this.hostname = parser.hostname;
+    //this.port = parser.port;
```
nburt, line 60: these needed?

drwestco: They're currently not used. I originally had them in for completeness, but didn't want to think about managing conflicts if they were written to.

PR #662, opened 2020-04-22
Make the search section sticky at top of page, except on mobile.

sikakkar:
image


drwestco:
Can the dark blue portion scroll so just the input controls are pinned? The big "let's get.." text takes up quite a bit of space.

sikakkar:
@drwestco had to do a bit of refactoring, but I think I got it.

sikakkar:
Ah, good catch @nburt! 

I'm not sure there's a good way to accomplish all the above goals (although there's a good chance I'm just missing an obvious answer):
- To get the thing to have position sticky, it has to be within the same container as the stuff it's sticking relative to (ie the instructions at the top and the list)
- Which means the map has to go in a different container, which puts it at the bottom in mobile

Questions for y'all:
1. Am I missing a way to do this? Maybe some flexbox cleverness that I didn't think of?
2. If not, are we okay with the top part "Let's get our PPE ..." also sticking to the top? Then we can go back to the old way of a flexbox with direction: row

nburt:
> Ah, good catch @nburt!
> 
> I'm not sure there's a good way to accomplish all the above goals (although there's a good chance I'm just missing an obvious answer):
> 
> * To get the thing to have position sticky, it has to be within the same container as the stuff it's sticking relative to (ie the instructions at the top and the list)
> * Which means the map has to go in a different container, which puts it at the bottom in mobile
> 
> Questions for y'all:
> 
> 1. Am I missing a way to do this? Maybe some flexbox cleverness that I didn't think of?
> 2. If not, are we okay with the top part "Let's get our PPE ..." also sticking to the top? Then we can go back to the old way of a flexbox with direction: row

Do you mind asking this in #ux-design? I think that I'm OK with 2 for now and then we should have real v2 mobile designs soonish and we can follow up then.

sikakkar:
@drwestco @nburt I think I found a way to make it so that not all the big font sticks at the top AND the map shows up in the right place on mobile!

views/index.handlebars
```diff
@@ -8,8 +8,8 @@
   {{> top_nav }}
 
   <div class="row flex-lg-row no-gutters">
-    <div class="col-12 col-lg-6">
-      <div class="bg-white search-filter-container">
+    <div class="col-12 col-lg-6 search-filter-container">
+      <div class="bg-white ">
```
saikatbhadra, line 12: nit - extra space

saikatbhadra:
Hey @sikakkar, I don't see any change in behavior, what should I be looking for here? The header still scrolls for me when I test the heroku deploy

drwestco:
Can you have it clip the instructional text too? Seems strange to have it stick here:
image

sikakkar:
@saikatbhadra what screen size are you using? I set it to this to restrict it to desktop:
`@media screen and ( min-width: 992px ) and ( min-height: 768px )`

sikakkar:
@drwestco I can, but unfortunately that messes up the placement on bigger screens where the text takes up 1 line instead of two
image

drwestco:
Yeah, it may be impossible to get it exactly right with just CSS. Might have to add some script to measure the header. Not sure what's worth the effort, though.

saikatbhadra:
Ah ok. @sikakkar I'm using a macbook air and on the default size and my height is 689 pixels. You might want to loosen up the height restriction

sass/style.css
```diff
@@ -48,11 +48,15 @@ a {
   color: white;
 }
 
-.search-filter-container {
-  position: sticky;
-  top: 58px;
-  z-index: 1;
-  background: white;
+@media screen and ( min-width: 992px ) and ( min-height: 768px )
```
saikatbhadra, line 51: Just tested in full screen mode, looks good but can you loosen the min-height a bit so more desktop users (esp ones with lower resolutions) get this effect?

saikatbhadra: From GA,

Here are most common desktop resolutions (window height will be smaller):
1920x1080
1440x900
1366x768
1536x864
1280x800

Here are mobile resolutions:
375x812
414x896
375x667
375x667

saikatbhadra: So looks like you can be more aggressive on the width side and less on the height?

sikakkar:
@saikatbhadra adjusted the media queries for width and height!

PR #679, opened 2020-04-23
Typography and spacing

tinabeans:


tinabeans:
hmm I'm going to see if I can fix these merge conflicts and re-submit

tinabeans:
Okay I think I fixed the merge conflicts. Who would like to review and merge? :)

sass/style.css
```diff
@@ -6,16 +6,27 @@ html {
...
-  color: #222;
-  font: normal normal normal 1rem/1.65 'Roboto', 'Helvetica', sans-serif;
+  color: #333;
+  font: normal normal normal 1rem/1.65 'Open Sans', 'Helvetica', sans-serif;
   height: 100%;
   margin: 0;
   padding: 0;
   text-rendering: optimizeLegibility;
+  font-size: 14px;
+  line-height: 1.55;
```
drwestco, line 16: You have 1rem/1.65 in the font property. Can combine here.

sass/style.css
```diff
@@ -45,6 +57,33 @@ a {
   background: white;
 }
 
+.search-filter-container .bg-dark-blue {
+  padding: 2rem !important;
+}
+
+.search-filter-container h4 {
+  font-weight: 700;
+  font-size: 22px;
```
drwestco, line 66: Should use rem units throughout.

sass/style.css
```diff
@@ -45,6 +57,33 @@ a {
...
+}
+
+.search-filter-container p {
+  font-size: 16px;
+}
+
+.search-filter-container .input-group-text {
+  padding-right: .7rem !important;
```
drwestco, line 74: Would be better to remove the p-2 class from the element and specify the complete padding here, instead of having an !important battle.

sass/style.css
```diff
@@ -45,6 +57,33 @@ a {
...
+}
+
+.search-filter-container .form-control {
+  height: auto;
+  font-size: 14px;
+}
+
+/* even out the spacing between filters... */
+.filters-container > *:first-child {
```
drwestco, line 83: Would be better to add a class to that first col element and use that to set padding, instead of this selector + !important

sass/style.css
```diff
@@ -181,12 +220,34 @@ h3.city {
   text-transform: uppercase;
 }
 
+.locations-container {
+  padding: 0 !important;
```
drwestco, line 224: You can remove the px-2 and m-2 classes from the element, and eliminate the need for !important.

sass/style.css
```diff
@@ -181,12 +220,34 @@ h3.city {
...
+  margin: 0 !important;
+}
+
 .locations-list-header {
+  padding: 0 !important;
```
drwestco, line 229: Can remove the px, pt, pb, mb, mt, and mx classes from the header element.

sass/style.css
```diff
@@ -181,12 +220,34 @@ h3.city {
...
+  margin: 0 !important;
   border-bottom: 1px solid #D4D4D4;
 }
 
+.locations-list-header > * {
+  padding: 1rem 2rem;
+  white-space:
```
drwestco, line 236: Missing value for this property

sass/style.css
```diff
@@ -181,12 +220,34 @@ h3.city {
...
+  margin: 0 !important;
   border-bottom: 1px solid #D4D4D4;
 }
 
+.locations-list-header > * {
+  padding: 1rem 2rem;
+  white-space: 
+}
+
+.locations-list-header #list-stats {
```
drwestco, line 239: IDs are unique, so no need for the class scope.

sass/style.css
```diff
@@ -181,12 +220,34 @@ h3.city {
...
+  font-size: 16px;
+  font-weight: 600;
+}
+
 .location {
   border-bottom: 1px solid #D4D4D4;
+  padding: 1.5rem 2rem !important;
```
drwestco, line 246: Can eliminate the py* class from the location element

sass/style.css
```diff
@@ -181,12 +220,34 @@ h3.city {
...
+}
+
+.location .font-weight-bold {
+  font-weight: 600 !important;
```
drwestco, line 250: Get rid of the font-weight-bold class on the element, add a semantic class instead and style that.

sass/style.css
```diff
@@ -358,19 +419,33 @@ h3.city {
...
+.navbar.ftm .navbar-brand,
+.navbar.ftm .add-donation-site-form {
+  display: flex;
+  min-height: 100%;
+  align-items: center;
+}
+
 .navbar.ftm .navbar-brand {
   flex-direction: row;
+  align-items: center;
```
drwestco, line 439: You already added align-items in the previous CSS block.

sass/style.css
```diff
@@ -451,6 +556,32 @@ h3.city {
...
+  .navbar.ftm {
+    padding: 0;
+  }
+
+  .navbar.ftm .navbar-brand {
+    margin-left: .5rem;
+  }
+
+  .navbar.ftm .navbar-nav .nav-link {
+    text-align: center;
```
drwestco, line 569: Centered items look bad in mobile, especially with the dropdown caret being included:

image

Looks especially bad for the share dropdown. The icons should be aligned:

image

Extra ick:

image

linderr:
@tinabeans girlie, you have comments <3

sass/style.css
tinabeans: Hm... we thought to center these links because accessibility would be better if a person doesn't have to reach all the way across to the left of the screen with their thumb, if they are right handed and have small and/or arthritic hands.

drwestco: That shouldn't be a problem. The full width is clickable.

tinabeans: I know, but our users don't know :) People naturally tap/click on what they can see, not what's invisible (ref: years of user testing folks from all walks of life)

tinabeans:
The test deployment failed so I'm going to try closing and reopening to trigger a new build.

sass/style.css
tinabeans: Lol I started fixing this, then noticed theres px sizes everywhere from before, so imma just leave it until later when I can transition this file over to using SASS vars

sass/style.css
tinabeans: fixed redundancy!

sass/style.css
tinabeans: Done!

sass/style.css
drwestco: Hmm. maybe there's some affordance we can use to indicate the whole row is clickable, instead of introducing the weirdness of misaligned icons?

tinabeans:
@drwestco I've decided to go ahead and merge so the fixes can get out to people, then I will do a big style fixes/cleanup in a separate branch with all of your comments in mind! Appreciate you for taking the time to point out all the ways this could be better.

sass/style.css
tinabeans: I hear you and we'll keep iterating on the design, but for now I'm going to go ahead and merge! Don't let perfect be the enemy of the good :)

PR #690, opened 2020-04-24
Added an email contact form that sends email to an encrypted recipient via a mail relay

raindrift:
This is the first version of the email contact frontend. The backend code lives at https://github.com/raindrift/maskmailer and runs at maskmailer.herokuapp.com

This adds a link to each organization in the list that pops a modal with an email contact form. Messages are relayed via an encrypted address, so that the recipient's address/identity are not revealed to the sender. A captcha helps to prevent abuse.

It's still pretty lightweight, but this should be enough to get it running and then we can add more features as we need them.

Two new feature flags (url GET params) are introduced here:

`show-contact`, when true, enables the contact form for listings that have an address
`fake-contact`, sets all contact addresses to my email, for testing


client/locations-list-map.js
```diff
@@ -1146,6 +1171,65 @@ function loadMapScript(searchParams, data, filters) {
...
+      }
+    ).done(function(result) {
+      $('.contact-form').css('display', 'none');
+      $('.contact-success').css('display', 'block');
+      $('#send-message').prop('disabled', false);
+      grecaptcha.reset();
+
+      setTimeout(function() {
+        $('#contactModal').modal('hide');
+      }, 5000);
```
ngottlieb, line 1217: why is this on a timeout?

client/locations-list-map.js
```diff
@@ -66,6 +66,8 @@ let gOpenInfoWindows = [];
 let gLocationsListEntries = [];
 let gLastLocationRendered = -1;
 
+const url = new URL(window.location);
```
ngottlieb, line 69: if you're putting `url` as a global here, can you also refactor the other places we're using it in this module?

views/partials/contact_form.handlebars
```diff
@@ -0,0 +1,39 @@
...
+            <div class="form-group">
+              <input type="text" class="form-control i18n" id="sender-email" placeholder="Your Email" data-i18n="[placeholder]ftm-email-form-email-placeholder">
+            </div>
+            <div class="form-group">
+              <input type="text" class="form-control i18n" id="message-subject" placeholder="Subject" data-i18n="[placeholder]ftm-email-form-subject-placeholder">
+            </div>
+            <div class="form-group">
+              <textarea class="form-control i18n" id="message-text" placeholder="Your Message" data-i18n="[placeholder]ftm-email-form-message-placeholder"></textarea>
+            </div>
+            <div class="g-recaptcha" data-sitekey="6Lc4zOgUAAAAAEVpMkSlTGuS81zHS34yNTMbf16r" data-size="compact"></div>
```
ngottlieb, line 25: can you make `sitekey` an env variable so we can use our own on dev?

client/locations-list-map.js
ngottlieb: possibly just assign `searchParams` as a global instead (and do the same)?

public/give.css
```diff
@@ -585,3 +585,23 @@ h3.city {
     margin-bottom: 10px;
   }
 }
+<<<<<<< HEAD:public/give.css
```
ngottlieb, line 588: accidentally committed some git notices

client/locations-list-map.js
nburt: Also @drwestco added a `FtmUrl` class for browser compat, can you switch to that

client/locations-list-map.js
```diff
@@ -683,6 +685,18 @@ function getFlatFilteredEntries(data, filters) {
 
 function getEntryEl(entry) {
   if (!entry.domElem) {
+    // feature flag for email contact form
+    const searchParams = new URLSearchParams(url.search);
+    const showContact = searchParams.get('show-contact') == 'true';
+
+    // feature flag to insert fake contact info for testing
+    const fakeContact = searchParams.get('fake-contact') == 'true';
```
nburt, line 693: do you think it'd be helpful to pass in fakeContact as an encrypted email string for testing? We could send around a few encrypted emails to testers so they could see the full flow rather than everything getting sent to you.

client/locations-list-map.js
drwestco: eslint should have caught this. did it not run for this PR?

client/locations-list-map.js
raindrift: It replaces the modal content with a "your email has been sent" confirmation. But there's no need to leave that up indefinitely.

client/locations-list-map.js
raindrift: Yeah, I can clean this up a little more. :)

public/give.css
raindrift: oh. derp. :)

client/locations-list-map.js
raindrift: Yeah. I was thinking this might be worth doing. I'll go ahead and add it.

raindrift:
> A couple small notes, awesome work! Have you tested how the modal behaves in an embed? It looks fine on the `/give` page but may be a bit funky in a smaller iframe context, I'm not sure.

I gave it a try on mobile (simulated) and in a small window, and it worked pretty well in those places.

raindrift:
> Should we also add this to the marker infowindows?

Probably. I'm thinking of maybe doing that as a separate PR. But we could also add it to this one if you think it makes sense. The modal will probably work from there...

nburt:
> > Should we also add this to the marker infowindows?
> 
> Probably. I'm thinking of maybe doing that as a separate PR. But we could also add it to this one if you think it makes sense. The modal will probably work from there...

makes sense, let's add it in a separate pr after this is merged

Issue #694, opened 2020-04-24
[bug] Facility type dropdown missing


saikatbhadra:
index and give used to have a facility type dropdown so you could filter by facility type. That seems to be missing currently, which I believe is a bug.

![Screen Shot 2020-04-24 at 9 45 26 AM](https://user-images.githubusercontent.com/2250961/80236885-a68bcd00-8610-11ea-872a-cf49bc813f17.png)
![Screen Shot 2020-04-24 at 9 45 16 AM](https://user-images.githubusercontent.com/2250961/80236888-a7bcfa00-8610-11ea-8504-5fbc97659dac.png)

In case this is a user specific bug
Browsers: tested on chrome and firefox

linderr:
CC @tinabeans please check the design of the fix!

ngottlieb:
@linderr the design of the fix is identical to how it was before it disappeared (it didn't actually get removed and added, the parsing of organization type from our dataset got broken)

linderr:
Perfect!!! @ngottlieb

PR #695, opened 2020-04-24
fix entry.orgType back to entry.org_type

ngottlieb:
resolves https://github.com/findthemasks/findthemasks/issues/694, but investigating if I should change the data upstream instead


ngottlieb:
looks ok, I went through the AirBNB guide and it doesn't seem to feel strongly about property naming. @awong-dev -- would you prefer I change the data upstream or use the prop with the underscore naming (`org_type`)?

Issue #698, opened 2020-04-25
www.findthemasks.ca directs to US-site


linderr:
www.findthemasks.ca should redirect to www.findthemasks.com/ca :) 
***map view + wrong side of the border is dimmed***

ngottlieb:
I can't replicate...

ngottlieb:
OHHH I see. It does redirect to `www.findthemasks.com/ca` but it's still showing US sites despite having the right URL!

linderr:
On Chrome: Map loads US -centre with US pins highlighed x Canada dimmed

ngottlieb:
yes, the problem is that even if you go directly to https://findthemasks.com/ca it's showing US sites. It only matches https://findthemasks.com/ca/ (with a trailing slash)

ngottlieb:
same for other countries (obviously)

linderr:
lol. Is that the default? oi

linderr:
btw I really appreciate this education - it makes sense why it didn't work
(but why does it work like that.................)

ngottlieb:
it's a bug for sure, just a tricksy one

linderr:
Cannot change www.findthemasks.ca ---> www.findthemasks.com/ca**/** ?

ngottlieb:
well, we could do that, but it also should just work to do www.findthemasks.com/ca in my opinion

linderr:
I agree :) thanks for fixing!!! I see it’s fixed for all international /fr /gb etc. Closing

PR #699, opened 2020-04-25
fix country code parsing to allow countries without trailing slash

ngottlieb:
https://github.com/findthemasks/findthemasks/issues/698

### DIFF
--- a/client/getCountry.js
+++ b/client/getCountry.js
@@ -4,7 +4,7 @@ import countries from './countries.js';
 export function getFirstPathPart() {
   const url = new FtmUrl(window.location);
   const directories = url.pathname.split('/');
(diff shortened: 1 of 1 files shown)


client/getCountry.js
```diff
@@ -4,7 +4,7 @@ import countries from './countries.js';
 export function getFirstPathPart() {
   const url = new FtmUrl(window.location);
   const directories = url.pathname.split('/');
-  if (directories.length > 2) {
+  if (directories.length >= 2) {
     return directories[1];
   }
```
saikatbhadra, line 10: can you log an error on line 26, so we can discern when this happens again through airbrake reporting?

PR #704, opened 2020-04-25
Webpack everything

awong-dev:


awong-dev:
@drwestco Is there any chance you might be interested in picking this up?  Or at weighing in on using jquery via CDN vs webpack?

We keep getting a small stream of airbrake errors from jquery being blocked. This would fix that... But it has a binary size cost.

awong-dev:
stale. closing.

Issue #709, opened 2020-04-25
redirect /gb -> /uk, make uk the canonical uk country code


nburt:

nburt:
https://github.com/findthemasks/findthemasks/pull/715

PR #715, opened 2020-04-26
Finishes #709 add redirect from gb -> uk and make uk the canonical code

nburt:
@awong-dev I haven't done anything with the data.json creation before but I think this will work?

Drew who is heading up the UK now confirmed that uk is a more inclusive country code / url over using gb. Per ISO standards uk is also accepted so this seems like the right move. I'm maintaining a gb redirect for backwards compatibility in case someone is linking to /gb but could be convinced to remove that.

I haven't run the firebase function yet, will wait for review first. This means the data-uk.json file isn't created yet (uk map won't work)


server.js
```diff
@@ -267,8 +267,22 @@ app.use('/data(-:countryCode)?.csv', createProxyMiddleware({
   changeOrigin: true,
 }));
 
-app.use('/', router);
+// redirect gb -> uk
+const gbUkRedirect = (req, res, next) => {
+  const { originalUrl } = req;
+
+  if (originalUrl.startsWith('/gb')) {
+    res.status(301).redirect(originalUrl.replace(/^\/gb/, '/uk'));
```
awong-dev, line 275: can we start with a 302? Just incase we somehow miss something with a regexp?

awong-dev:
looks good with one comment.

PR #716, opened 2020-04-26
Add back analytics to top nav social sharing

saikatbhadra:
The new top nav wasn't setting events to google analytics - this fixes that.


ngottlieb:
:shipit:

PR #727, opened 2020-04-27
enable email relay for users who have not opted out

nburt:


nburt:
I've already deployed the function to prod + run reloaddatasheet on each country so that the encrypted email field is stripped

linderr:
May we talk about this before merging??? Especially w/the data leads @4kbt @cchang-adi

4kbt:
Let's wait until the morning (see discussion in #dev) to merge. We want to be real sure that we don't antagonize recipients :).

functions/index.js
```diff
@@ -334,13 +338,25 @@ async function snapshotData(country) {
...
+    const contactFormOptOutIndex = orig_col_labels.findIndex(e => e === '_contact_form_opt_out');
+
     const trimmed_values = real_values.map((row, row_num) => {
       const result = [];
 
+      // Opt out of mail relay if
+      // country's sheet does not have opt out option
+      // OR opt out column has content in it
+      const hasOptedOutOfContact = contactFormOptOutIndex === -1
+        || (Boolean(row[contactFormOptOutIndex]) && row[contactFormOptOutIndex].trim().length > 0);
```
awong-dev, line 350: How can (Boolean(row[contactFormOptOutIndex]) be true
and row[contactFormOptOutIndex].trim().length be 0?

nburt: The index === -1 check might have fixed this... I can revisit but originally was checking to make sure that we weren't calling `trim().length` on undefined but we might not need that.

Ultimately I think this field should be explicit as an enum with 'Yes' / 'No'. We can migrate to that by:

1. Replacing the current opt out text with "Yes" but leaving the no's as blank
2. Update functions/index.js to check for presence + value === 'Yes'
3. Update form to write No answers too

PR #729, opened 2020-04-27
Findthemakers embed and other fixes

awong-dev:
Start generalizing our templates and data format to handle non-PPE requester data.


awong-dev:
Don't merge yet! This has filters commented out.

client/locations-list-map.js
```diff
@@ -79,31 +89,33 @@ const translatedFilterItems = (filterItems) => {
...
-        entry.accepting.split(/, (?![^(]*\))/).map((a) => a.trim()).forEach((i) => {
-          const filterKey = i.toLowerCase();
-          if (FILTER_ITEMS[filterKey] !== undefined && acceptedItems[filterKey] === undefined) {
-            acceptedItems[filterKey] = {
-              ...FILTER_ITEMS[filterKey],
-              value: filterKey,
-            };
-          }
-        });
+        if (entry.accepting) {
```
ngottlieb, line 108: do we need filters on makers? I'm sure you thought of this, but it'd be easy to add a conditional in here on `dataset` to generate filters from different columns. We could even define dataset properties somewhere (`datasets: { ppe: { filters: ['org_type', 'accepting'] }, makers: { filters: ['capabilities'] } }`) and use that throughout to generalize this code more.

awong-dev: I spent 2 hours last night hacking through filters for makers and gave up.

The answer is "yes", but I couldn't make it work so just wanted the list out first. :-/

client/locations-list-map.js
```diff
@@ -5,7 +5,7 @@ import MarkerClusterer from '@google/markerclustererplus';
 import toDataByLocation from './toDataByLocation.js';
 import countries from './countries.js';
 import { FILTER_ITEMS, ORG_TYPES, ENUM_MAPPINGS } from './formEnumLookups.js';
-import { getCountry, getFirstPathPart, isCountryPath } from './getCountry.js';
+import { getCountry } from './getCountry.js';
```
drwestco, line 8: Is this even needed? Can the script just trust body.dataset.countryCode?

client/locations-list-map.js
```diff
@@ -66,9 +65,18 @@ let gLastLocationRendered = -1;
 
 const searchParams = new FtmUrl(window.location.href).searchparams;
 
+// Converts a string from 'a,b,c' to 'a, b, c'
+function addSpaceAfterComma(str) {
+  if (str) {
+    return str.split(',').join(', ');
```
drwestco, line 71: Could use a regex in the call to split, so you don't keep adding spaces if the source string already has them.

client/locations-list-map.js
```diff
@@ -231,7 +260,53 @@ function multilineStringToNodes(input) {
   return returnedNodes.slice(0, -1);
 }
 
-function addMarkerContent(orgType, address, name, instructions, accepting, openBox, separator) {
+function createMakerMarkerContent(entry, separator) {
+  // Text to go into InfoWindow
+  const contentTags = separator ? [ce('h5', 'separator', ctn(entry.name))] : [ce('h5', null, ctn(entry.name))];
```
drwestco, line 265: Cleaner to put the conditional inside the call to ce, since the other two params are the same
```javascript
const contentTags = [ce('h5', separator ? 'separator' : null, ctn(entry.name))];

File: client/locations-list-map.js [hunk 5]
```diff
@@ -907,13 +1037,22 @@ function loadDataFile(url, dataToStore) {
   );
 }
 
+function getDatasetFilename(dataset) {
```
drwestco, line 1040: This should take both dataset and countryCode

```javascript
if (dataset === 'makers') { return '/data-makers.json'; }
else { return `/data-${countryCode}.json`; }

File: client/locations-list-map.js (hunk 2 above)
Comment (drwestco) [reply client/locations-list-map.js:8]: Just one call to getCountry when setting the map's initial view. I think that can use gCountryCode too.

File: server.js [hunk 6]
```diff
@@ -196,9 +198,22 @@ router.get(['/404', '/404.html'], (req, res) => {
...
+});
+
+router.get('/maker-form', (req, res) => {
+  res.redirect('https://airtable.com/shruH5B27UP3PqKgg');
 });
 
+// Recursively handle routes for makers overriding the dataset so the main
+// map functionality can be run in a different "mode" so to speak.
+router.use('/makers(/embed)?$', (req, res, next) => {
+  res.locals.dataset = 'makers';
```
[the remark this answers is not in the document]
awong-dev: I'll add a comment, but not required.  The top-level routing is setting it.

server.js
```diff
@@ -281,8 +296,34 @@ const gbUkRedirect = (req, res, next) => {
...
-app.use('/', router);
+
+const ALL_COUNTRIES = new Set(Object.keys(countries));
+
+app.use('/:countryCode', (req, res, next) => {
+  const lowerCased = req.params.countryCode.toLowerCase();
+
+  if (ALL_COUNTRIES.has(lowerCased)) {
+    res.locals.countryCode = req.params.countryCode;
+    res.locals.dataset = res.locals.countryCode;
```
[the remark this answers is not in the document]
awong-dev: I understand what you're saying now...  Cool. Done.

Also changed it to 'requester' instead of 'default' cause it's more semantically meaningful.

server.js
```diff
@@ -281,8 +296,34 @@ const gbUkRedirect = (req, res, next) => {
...
+
+    // Redirect to lower-cased path.
```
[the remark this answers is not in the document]
awong-dev: it's better in in terms of canonical URLs. More friendly to caches, seo, analytics, etc.

views/give.handlebars
```diff
@@ -12,7 +12,14 @@
 
           <div class="row" style="margin: 0px 0px 20px 0px;">
             <div class="col-xl-4" style="padding: 0px">
-              <h3 style="white-space: nowrap"><span class="i18n" data-i18n="ftm-map-of-sites">Donation sites</span> <span id="map-stats"></span></h3>
+              <h3 style="white-space: nowrap">
+                {{#if (eq dataset "makers")}}
+                  <span>Maker Groups</span>
```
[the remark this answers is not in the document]
awong-dev: Gonna do the i18n for makers in a later pass.

I don't understand the strategy for how to name all those things.

client/locations-list-map.js
awong-dev: Good point. Used gCountryCode.  I think getCountry() probably should be removed from everywhere now. Leaving for another CL.

server.js
```diff
@@ -99,15 +100,16 @@ router.get(['/give', '/give.html'], (req, res) => {
...
-router.get(['/makers'], (req, res) => {
-  res.render('makers', {
+router.get(['/embed'], (req, res) => {
+  const isMaker = res.locals.dataset === 'makers';
+  res.render('give', {
     version: herokuVersion,
+    layout: 'give',
     ogLocale: formatFbLocale(res.locals.locale),
-    ogTitle: res.locals.banana.i18n('ftm-makers-og-title'),
+    ogTitle: isMaker ? '#findthemasks | makers embed' : res.locals.banana.i18n('ftm-give-og-title'),
```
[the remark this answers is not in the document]
awong-dev: will do in another pass.

client/locations-list-map.js
awong-dev: Yeah that is much better. Was copy pasta code. Fixed.

client/locations-list-map.js
awong-dev: Done.

client/locations-list-map.js
awong-dev: Not as worried for now since it's a machine generated enum. Also it's html thus extra spaces collapse.

awong-dev:
If this passes travis and looks okay on heroku, I'll merge and we can fix other issues in a follow up

PR #736, opened 2020-04-27
Resolves #725: end ga event when email is sent through the relay

### LINKED ISSUE #725: Track email relay submission in GA
- sends
- errors?
- other?

nburt:


client/locations-list-map.js
```diff
@@ -1216,6 +1216,7 @@ function initContactModal() {
       $('.contact-success').css('display', 'block');
       $('#send-message').prop('disabled', false);
       window.grecaptcha.reset();
+      sendEvent('contactOrganization', 'sendEmail', $('#contactModal').find('.modal-title').val());
```
saikatbhadra, line 1219: can we do this in the back end (i.e. mail gun) rather than through the client side? The numbers will be more accurate on the server side.

On the client side I would instrument 3 events:
1. when the user clicks on sendEmail (click event)
2. when the request back (this event / or can be done through mailgun)

Also if you have time can you send airbrake error if there's a email failure? (either through client side or server side)

nburt: We were thinking of doing it client side so that the data is all in one place. We have mailgun too but this seemed like a good enough proxy.

I can update to the click event + done event. 

Airbrake errors are another thing... I suppose we should add it to the maskmailer repo. @raindrift do you have any error tracking for it?

PR #763, opened 2020-04-29
Fix embed collaboration string and shrink NoM logo

awong-dev:


awong-dev:
Basic PR app checks out. Will wait 15 mins before merge in case there are comments.

PR #767, opened 2020-04-30
Add in virtual host for findthemakers.com.

awong-dev:
Also split out the routing in server.js to make it more readable.


awong-dev:
After this is merged, we will get Stephen to fix the DNS to be a cname
instead of a redirect.

Then, yeah, one server will be powering two urls.

-Albert

PR #769, opened 2020-04-30
Add a little more bottom margin to map

sunnymui:
Text under makers map still a little close.  Up the margin a little more for more space.
Does the same for the donation site map, though there's an inline margin set on the form button overriding everything for that map that expands the margin even further

### DIFF
--- a/sass/give.scss
+++ b/sass/give.scss
@@ -275,7 +275,7 @@ h3.city {
 #map {
   height: 400px;
   width: 100%;
-  margin-bottom: .3ex;
+  margin-bottom: .5em;
 }
 
 #map .label {


nburt:
This looks good, do you think we should remove the margin-top on that button for the requesters embed?

Looks a little tight with the button but maybe we could increase the map margin-bottom from 0.5rem -> 1rem. Would be great to have consistent margin for both and not have the random button margin.

image

nburt:
this is 1rem margin, above is 0.5rem with button margin removed:

image

sunnymui:
Yeah we can do 1rem, would help keep things consistent.

Is that btn-sm or btn-primary class being used anywhere else? I think it needs a mobile breakpoint to have a small margin on top because the button gets a little close to the text on smaller screens.

nburt:
> Yeah we can do 1rem, would help keep things consistent.
> 
> Is that btn-sm or btn-primary class being used anywhere else? I think it needs a mobile breakpoint to have a small margin on top because the button gets a little close to the text on smaller screens.

We use getbootstrap.com for basic styling and those are from that. If you need to override that you can use bootstrap's margin utility classes or you can define a new class with the styles

sunnymui:
Is that html for the button being fetched from a json? I don't see it anywhere in the repo files

nburt:
yeah, i18n is a little confusing. Strings are stored in en.json. I just added you as a collaborator to the repo. Can you close this PR, update that string, and then open a new one from the main repo rather than the fork? This will allow our translation platform (Smartling) import the content for translation. It doesn't work from forks unfortunately.

This wiki page has more info on the process: https://github.com/findthemasks/findthemasks/wiki/Translation-Process

Also feel free to ping me on slack if you have any questions

PR #796, opened 2020-05-01
Css cleanup

tinabeans:


sass/give.scss
```diff
@@ -1,3 +1,5 @@
+@import 'colors';
+@import 'typography';
```
awong-dev, line 2: At some point should give.scss share more of style.scss do you think? Not asking a change in this PR. Just trying to understand direction.

sass/colors.scss
```diff
@@ -0,0 +1,9 @@
+$text-color: #333333;
```
nburt, line 1: 🎉

sass/give.scss
nburt: Let's aim to do that as part of the v2 embed work: https://github.com/findthemasks/findthemasks/issues/649

sass/style.scss
```diff
@@ -226,13 +228,13 @@ h3.city {
...
+  padding: 0;
+  margin: 0;
 }
 
 .locations-list-header {
-  padding: 0 !important;
-  margin: 0 !important;
+  padding: 0;
+  margin: 0;
   border-bottom: 1px solid #D4D4D4;
```
nburt, line 238: Can you make this a variable too?

sass/style.scss
```diff
@@ -314,19 +317,26 @@ h3.city {
   border-top: 1px solid darkgray;
```
nburt, line 317: this gray should mark the border above

nburt:
Added another color var and merged in master

PR #812, opened 2020-05-02
first pass at ux2 whoweare page

### LINKED ISSUE #801: Update About Us subpage: content + UX/UI
**Design only** as per two-column "About-Desktop" frame on Figma: https://www.figma.com/file/qbibGHgcV1LBcpvd2EoHSM/FindTheMasks?node-id=287%3A4547

**Latest Content** page 9-12
https://docs.google.com/document/d/1b9lj2j3vqJrZgqfozbZ2reSO6xnC_a00VtKO-xxUB7g/edit#
- text sections indicated which should go on the right column 
- right column is region-specific

image

ngottlieb:
resolves #801


linderr:
@ngottlieb awesome!! I think as I work to back-fill people's preferred links and the right-hand side becomes blue, it'll be more visually distinct from the left! Great work!

linderr:
How do you propose I help to conserve the translation quota as we move forward with updating names and links?

I'm thinking once we get the base set up, i'll just do one name/link content update a week?

ngottlieb:
None of the names are in a translated text block so it's not an issue.

linderr:
Yessssss great.

sass/style.scss
```diff
@@ -1,6 +1,6 @@
-@import 'colors';
-@import 'typography';
-@import 'icons.data.svg';
+@import "colors";
+@import "typography";
+@import "icons.data.svg";
```
ngottlieb, line 4: oops sorry for all the linting changes on the `style.scss` file, was trying to avoid. But I think they are fine -- can re-work the commits to exclude formatting changes if preferred

views/whoweare.handlebars
```diff
@@ -19,7 +19,7 @@
           This site was launched on <a href="https://twitter.com/arrkay/status/1240704610186092546">March 19th</a>
           by <a href="https://www.linkedin.com/in/rachelpopkin/">Rachel Popkin</a> and has been built and improved
           upon by many fantastic volunteers worldwide. <br /><br />
-          <strong>The team is made up entirely of volunteers</strong> and includes physicians, scientists, physicists, epidemiologists,
+          <b>The team is made up entirely of volunteers</b> and includes physicians, scientists, physicists, epidemiologists,
```
linderr, line 22: How do I make something **bold**
I tried `<strong>` and `<b>` tags :(

linderr:
@ngottlieb Conflicting files applicationRoutes.js ??

public/i18n/en.json
```diff
@@ -179,7 +179,7 @@
   "ftm-requesters-count": "$1 requesters",
   "ftm-about-us-title": "About us",
   "ftm-about-us-what": "What is #findthemasks?",
-  "ftm-about-us-what-details": "<strong>#findthemasks is a global mapping tool and interactive directory for PPE needs.</strong> Our mission is to enable individuals, communities, makerspace networks, and large-scale manufacturers to engage and assist during the COVID-19 pandemic by providing our frontline healthcare and other essential workers with the personal protective equipment they need to protect themselves.",
+  "ftm-about-us-what-details": "<strong> #findthemasks is a global mapping tool and interactive directory for PPE needs.</strong> Our mission is to enable individuals, communities, makerspace networks, and large-scale manufacturers to engage and assist during the COVID-19 pandemic by providing our frontline healthcare and other essential workers with the personal protective equipment they need to protect themselves.",
```
linderr, line 182: why do these <strong> tags show up in the herokku site instead of bolding the text?

ngottlieb: Hahah -- when we have HTML in translation text that needs to be specified where the translation text is rendered. I'll add it

applicationRoutes.js
```diff
@@ -3,9 +3,7 @@
 const express = require('express');
 const formatFbLocale = require('./utils/formatFbLocale');
 const getDonationFormUrl = require('./viewHelpers/getDonationFormUrl.js');
-const localizeContactInfo = require('./viewHelpers/localizeContactInfo.js');
-const selectMaskMatchPartialPath = require('./viewHelpers/selectMaskMatchPartialPath');
-const selectLargeDonationSitesPartialPath = require('./viewHelpers/selectLargeDonationSitesPartialPath');
+const localizePartialPath = require('./viewHelpers/localizePartialPath');
```
nburt, line 6: 💯

sass/style.scss
nburt: fine by me

sass/style.scss
```diff
@@ -112,23 +115,22 @@ a {
...
-  width: 24px;
-  height: 24px;
+  width: 25px;
+  height: 25px;
 }
 
 .member-of {
   font-style: italic;
-  font-size: .8125rem;
+  font-size: 0.8125rem;
```
nburt, line 132: should this be $font-size-small?

sass/style.scss
```diff
@@ -210,7 +212,7 @@ a {
 h2.state {
   background: #eee;
   color: #000;
-  font: 800 24px 'Roboto';
+  font: 800 24px "Roboto";
```
nburt, line 215: noticing nits cause of the fomatting changes. We don't use roboto anymore and I don't think we use this class either. We can do a separate styles cleanup pass

nburt:
Is there anything blocking merge or can we do that and then do any other followups after?

linderr:
+1 for merging I’ll add link outs in small batches as I get them in a new PR, exciteddd!!

public/i18n/en.json
linderr: Ohhhh is that the [html] necessary in .handlebars? Ok! I get it now - previously I just labelled it for a href= links but now I understand it’s a label for HTML language coooooool!! I’ll be more careful next time, thanks!!

nburt: yeah basically if any strings in en.json contain any HTML tags in them, you need to add [html] before the key so that it is inserted as a string rather than text

linderr: That makes so much sense! Ok, got it for next time

sass/style.scss
ngottlieb: Yeah I'll create another issue for that

linderr:
wooOOOoooooO!!!

PR #824, opened 2020-05-03
Fixes #768: make filters unique by display name

### LINKED ISSUE #768: Homemade Masks shows up twice in filter list?
Screen Shot 2020-04-30 at 12 15 11 AM

nburt:


nburt:
We've adjusted the form over time so this used to be just "Homemade masks" and then was updated: 

image

So this is the case of handling form values changing over time. We could try and fix a level up and I'm open to that before merging but seemed like we might end up just needing this anyways

ngottlieb:
K -- I'm fine with it here :+1:

PR #834, opened 2020-05-03
server render nav

nburt:
This is just the locales. I haven't really ever used handlebars / view helpers / express middleware in this way. Would love some feedback on the approach and then will finish the rest of the navbar


client/locales.js
```diff
@@ -1,4 +1,4 @@
-export default [
+module.exports = [
```
nburt, line 1: will move out of client before merging

client/render-nav.js
```diff
@@ -1,41 +1,16 @@
...
-      element.className = 'dropdown-item';
-      const currentUrl = new FtmUrl(window.location.href);
-      currentUrl.searchparams.locale = locale.localeCode;
-      element.setAttribute('href', currentUrl.toString());
-      element.textContent = $.i18n(locale.i18nString);
-      element.addEventListener('click', () => sendEvent('i18n', 'set-locale', locale.localeCode));
-      localeDropdownItems.appendChild(element);
-    });
+  if (countryDropdownLink && countryDropdownItems) {
+    // element.addEventListener('click', () => sendEvent('i18n', 'set-locale', locale.localeCode));
```
nburt, line 13: need to figure out how to add this back

nburt:
I went a little rogue here on the linting but was getting some browser compat plugin errors in server.js so I split that out.

We now have 3 eslint files. I can tackle merging the functions config with the root one but there were tons of errors and I didn't want to add more to this PR.

server.js
```diff
@@ -1,21 +1,55 @@
 const express = require('express');
 const expressHandlebars = require('express-handlebars');
 const vhost = require('vhost');
-const setBananaI18n = require('./middleware/setBananaI18n.js');
 const rootRoutes = require('./rootRoutes');
 require('dotenv').config();
 require('handlebars-helpers')();
 
+const handlebars = expressHandlebars.create({
```
ngottlieb, line 8: I probably don't have any more experience with using Handlebars and Express this way than you do, but what about moving these helpers to `/viewHelpers` (like our other helpers)? The difference, of course, is that you're applying these to the engine, not to each route, but what you could do is define all three in one `viewHelpers/navHelpers.js`, then

    import { createCountryDropdownHref, createLocaleDropdownHref, createNavbarItemHref } from './viewHelpers/navHelpers.js'

then

    const handlebars = expressHandlebars.create({
      helpers: {
        createCountryDropdownHref,
        createLocaleDropdownHref,
        createNavbarItemHref
      }
    })

feel free to ignore, but I *think* these functions are the architectural equivalent of hte other ones we're storing in `/viewHelpers`, so they should probably be in the same place?

ngottlieb:
Oops one more comment -- not that critical because Makers only works for USA but, the "Find the Masks" / "Find the Makers" link isn't using localization logic

nburt:
i18n ticket existshttps://github.com/findthemasks/findthemasks/issues/836

Issue #740, opened 2020-04-28
Add ability to link to an FAQ with id hash and have it default open


nburt:
As part of updating contact email formatting we want to link to a specific FAQ section from the email footer similar to how craigslist does it: https://github.com/findthemasks/findthemasks/issues/723

For that to work we need the ability to link to a specific question / answer. I think this could be useful elsewhere

linderr:
Makers want this as well to jump straight to the Makers FAQ section

saikatbhadra:
You can link to any element of the faq that has id tag associated with it. You can check it out by trying out this link: https://findthemasks.com/faq#ftm-faq-how-to-help

PR #838, opened 2020-05-04
Setup FAQ for linking

saikatbhadra:
- Sets up id for each FAQ question so #hyperlinks work. Sets the id on the `<dt>` but perhaps we should do it on the `<dd>`?
- Adds analytic event when a user clicks on a link to a question or expands a question

### DIFF
--- a/views/faq.handlebars
+++ b/views/faq.handlebars
@@ -10,18 +10,17 @@
       <section>
         <h4 class="i18n faq-header" data-i18n="ftm-faq-basic-faqs-header">Basic FAQs</h4>
         <dl class="faq-question">
-          <dt class="btn-link open">
+          <dt class="btn-link" id="ftm-faq-what-is-ppe">
             <h5 class="i18n" data-i18n="ftm-faq-what-is-ppe">What is PPE?</h5>
           </dt>
-          <dd class="collapse show">
+          <dd class="collapse">
             <p class="i18n" data-i18n="ftm-faq-what-is-ppe-answer">Personal protective
               equipment (PPE)
               is protective clothing, goggles, or other equipment designed to protect the wearer from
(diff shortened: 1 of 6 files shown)


saikatbhadra:
Closes: https://github.com/findthemasks/findthemasks/issues/740

views/faq.handlebars
```diff
@@ -233,11 +230,10 @@
           </dd>
         </dl>
       </section>
-
       <section id="section-makers">
         <h4 class="i18n faq-header" data-i18n="ftm-faq-makerspace-header">Makerspace Community FAQs</h4>
```
linderr, line 234: Can we adopt this for the header class?

linderr:
This is very cool. Thanks for doing this @saikatbhadra ! I see you enabling us to link to exact questions.
Would it be possible to adopt this for the header class? We also want to link people sections. 
Ideally for the Makers it would be https://findthemakers.com/makers/faq#ftm-faq-makerspace-header

saikatbhadra:
Any id works on the html. Looking at the makers section it looks like it currently has an id of `section-makers` so if you link to https://findthemasks.com/faq#section-makers then it will go to the makers FAQs.

linderr:
That’s great! If you have time to write a Wiki for linking using that for the non-devs like me to refer to when interacting with collaborators/partners or on socials. I can also work on that later tonight & ask for your help to edit it?

saikatbhadra:
In this case, we aren't doing anything special, we are just following standard html method for how to link to different parts of a page. Here are two articles that explain how it works, you could link to them on the wiki:

https://blog.hubspot.com/marketing/jump-link-same-page
https://stackoverflow.com/questions/2835140/how-do-i-link-to-part-of-a-page-hash

PR #842, opened 2020-05-05
convert embed to UX2

### LINKED ISSUE #649: [v2-design] Carry v2 design flourish to embed page - Designs DEV READY on Figma
give.html should look similar to index.html with the caveat that since this page is being embedded on other sites, it should deal with smaller window sizes. Getusppe restricts height to 750px.

Some features we need to continue to support
- [x] any of the embed query params (turn filters on/off, center map to location, center map to state, etc

Test on the following sites:
- [x] https://getusppe.org/give/ (embed is in middle of page)
- [x] https://theppedrive.com/ppe-drive-listing
- [x] https://resource19.org/pages/search.html
- [x] https://lafightingcovid.com/

ngottlieb:
Trying to do some css cleanup along the way here, hopefully this doesn't take me too much longer because my CSS changes will be a bit divergent if anyone else does CSS work between now and merge.

TODO:

- [x] basic styling for the table view
- [x] determine importance of footer, figure out how to incorporate if necessary
- [x] consider collapse logic further, see Slack for details, if collapse is coming to "main tab" code soon (in mobile design) then not worth doing here
- [x] continue the CSS cleanup started in this branch (at least for `give.scss`)

resolves #649


ngottlieb:
@linderr I got the collapse logic to work fairly unobtrusively. Only remaining open question is the importance of the footer, and given the hypothetical importance of the footer, what the max height we want for map/table interface is?

linderr:
@ngottlieb the footer is important. We see on Google Analytics it drives traffic back to the main site :)

linderr:
Does the ?hide-list=true result in the map-only view? https://www.figma.com/file/qbibGHgcV1LBcpvd2EoHSM/FindTheMasks?node-id=980%3A0

ngottlieb:
Yeah, I dealt with the footer. And yes

nburt:
I embedded on getusppe using the heroku url - looks like the list is hidden but map doesn't expand to 100% width:

image

client/locations-list-map.js
```diff
@@ -1442,6 +1448,25 @@ const applyFilterParams = ((params, filterSet) => {
   });
 });
 
+const initializeEmbedLocationCollapse = () => {
```
nburt, line 1451: thoughts on trying to remove jQuery dependency at some point? I think we were trying to get there but it might just not be worth it at this point given that we use bootstrap's popover / modal now

nburt:
A few other UI tweaks. 

Top margin / centering of loading spinner:

image

Right align org type on desktop:

image

When hide-map=true locations should be expanded:
image

client/locations-list-map.js
ngottlieb: Hahah, yeah, I try not to use jQuery in my "real" life these days but I'm not sure it's worth us trying to escape it for this project, esp. given your point about Bootstrap JS

ngottlieb:
@nburt I can't replicate your issue on getusppe, it looks right for me.

Fixed the rest of your issues except the multi-line org type display when the main hospital name is too long. I already banged my head against the wall of the flex-box trying to solve that earlier and have not yet succeeded. I'll try again, hahah

ngottlieb:
OK I think all issues resolved.

ngottlieb:
Happy to have any additional feedback on this since it's a big release that impacts a number of our partners but I think it's ready to merge. cc: @linderr, @saikatbhadra ?

linderr:
How do I view as other other sites as you guys did for GUPPE? I can go through all embed pages

saikatbhadra:
It’s a bit technical. You go to the embedded site you want to test and then
you find the iframe in the chrome debugger and modify the HTML to point to
the heroku site instead of our live site.
-- 
-Saikat
saikat.bhadra@gmail.com
(832) 654 1735
https://www.linkedin.com/in/saikatbhadra

linderr:
ohhhhhhhh I see

linderr:
T_T why am I so slow with this omg

linderr:
thehomefront.ca
www.theppedrive.com
https://maskaherobc.ca/
https://maskaheroontario.ca/
www.covid19central.ca
https://www.redmooncosplaysolutions.com/
https://kba-nyc.com/mask-distribution

linderr:
When ?hide-list=true, it still hides the filter menus too
image

ngottlieb:
That's the behavior on the existing embed

linderr:
Correct but an artifact of v1 main + v1 embed.

v1 had search bar on top and filter menus below the map.
v2 embed keep both search + filter categories on the map-only view:
image

linderr:
?hide-list=true should only hide the left-hand-side-collapsable-list-thingy, and exclude the two filter menus

saikatbhadra:
I'm still seeing footer issues:
Screen Shot 2020-05-05 at 6 21 39 PM

saikatbhadra:
I'm also unable to get to the footer on mobile (I think its do to multiple scrollsbars):
Screen Shot 2020-05-05 at 6 31 53 PM

ngottlieb:
@saikatbhadra what browser? Looks ok to me in Brave (Chromium) (desktop view).

Re: mobile...I'll play around with it and see what I can do. I think it's probably ok if the footer is inaccessible? It wasn't in the design mockup for mobile anyway.

linderr:
There was a little blurb under both map-only and list+map mobile mockups but if the footer is unruly let's just move the info? I think we have some autonomy here to deviate from the designs if something doesn't work in reality 😃 !

It was the right call not to do the page number thing suggested @ngottlieb . I also like the idea to shorten the height of the map as per mockup for mobile. 
image

saikatbhadra:
**List Of Blocking Issues**
1. Change the footer so it shows so its visible above the fold for users with smaller phones, here's a screenshot of an iphone 8 and its not visible:
Screen Shot 2020-05-05 at 9 28 29 PM
2. Figure out how to satisfy the requirement "as a mobile user, I can see the footer without scrolling" Some possible solutions:
- Move footer content into header
- Move footer placement between map view and list view
- Make footer sticky
- Reduce height of map in certain resolutions (basically same as #1) - this is the option I am leaning to since it's the easiest to implement 
3. Fix  the hide-map=true/false param. It's breaking when hide-map=false. Try it on this link: https://findthemasks-feature-ux-sc3sqy.herokuapp.com/give.html?state=ca&hide-map=false&hide-filters=false (it is a param used by KTLA)
4. Fix broken footer on this url: https://findthemasks-feature-ux-sc3sqy.herokuapp.com/give.html?state=ca&hide-filters=false&hide-list=false&hide-search=true&locale=es. I think it breaks on some other scenarios, but this one I can repro easily. It looks like the mapview is flowing down past the list when it shouldn't and the footer is hovering over it. Screenshot from 2015 macbook air (low resolution) on chrome here:
Screen Shot 2020-05-05 at 10 37 28 PM

linderr:
collaborator context:
We have teased the Figma mockups as potential appearances of embed v2. We presented the embed list-view as a new option which people thought was cool. Some collaborators still wanted to keep the map-only view b/c it looks cleaner on their webpage so we still have to support hide-list

ngottlieb:
Ah, hide-map and hide-list are evaluating as true whenever any string is passed. I sort of assumed no one used things like "hide-map=false" but I can process that...

ngottlieb:
@saikatbhadra most of the footer issues you're pointing to are actually `hide-xxxx` issues manifesting in the form of footer issues.

linderr:
Re: no one uses hide-map=false
You would think that but then there is KTLA :)

ngottlieb:
@saikatbhadra can you test again on mobile? I'm trying it now, footer doesn't show up above the fold on my phone, but TBH that seems pretty much impossible. It *is* easy to navigate to (on Android at least) because the "scroll" is so much shorter now.

saikatbhadra:
Still seeing the footer issue when I test on resource19  @ngottlieb (https://resource19.org/pages/search.html):
Screen Shot 2020-05-06 at 10 10 25 AM

PR #846, opened 2020-05-05
Canva to Partners - what is the logic?

linderr:


views/partners.handlebars
```diff
@@ -32,6 +32,11 @@
             <img src="/images/logos-Sponsors/Airbrake.svg" alt="Airbrake Logo">
           </a>
         </div>
+        <div class="partner-logos partner-logos-small">
+          <a href="https://www.canva.com/templates/" target="_blank" class="partner-logo-container">
+            <img id="img-logo-canva" src="/images/logos-Sponsors/Canva.png" alt="Canva Logo">
+          </a>
+        </div>
```
linderr, line 39: @ngottlieb is this better logic than adding a id tag in CSS for this one logo?
Because it's a circle, displayed on 100% as per partner-logos-large it just looks MASSIVE to me on the Sponsor list.

I tried the #id CSS thingy but now I am thinking to adopt other styling is probably more logical in the future if we make any changes, there won't be this one random logo with it's own styling. I would like it to just be the size of the Media icons but then the part I don't like becomes a the `<div>` class moves it to the next row. Would love to learn how you would solve this. I'm okay with putting the #id in CSS but it doesn't feel preemptively "logical"

nburt: Id or class is better I think. This way also bumps it out onto its own row rather than having 3 per line on desktop.

linderr: Sounds good - I'll do that now. I did that in the other pull request which I closed but decided maybe it's not protective of future changes and changed my mind.
If we end up changing style in the future we can revisit :)

PR #875, opened 2020-05-07
Zoom out & recentre the US default view

linderr:


linderr:
Travis didn't like my comment 👍

PR #880, opened 2020-05-07
Add OSMS data into makers dataset.

awong-dev:
Currently it's a csv so just parsing locally, doing the geocode
and then appending the dataset. This also adds a proper rate limit
to the geocoding API from google maps so we don't trigger the abuse
limit.

### DIFF
--- a/functions/airtable-connector.js
+++ b/functions/airtable-connector.js
@@ -1,6 +1,10 @@
 const Airtable = require('airtable');
 const functions = require('firebase-functions');
+const csv = require('csv-parse');
+const fs = require('fs');
 const zipcodes = require('zipcodes');
+const { geocodeAddress } = require('./geocode.js');
+const { readAirtableSharedView, parseAirtableData } = require('./airtable-shared-view.js');
 
 const FIELDS = [
       "approved", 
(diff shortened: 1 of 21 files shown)


functions/airtable-connector.js
```diff
@@ -70,7 +73,7 @@ function toDataJson(entries) {
   };
 }
 
-function loadMakerData(admin, req, res) {
+function loadFromAirtable(static_data, admin, req, res) {
   const base = new Airtable({apiKey: 'key7sP7fMv2drjUnJ'}).base('appmMuprpuCFw4wCd');
```
nburt, line 77: config vars -> google cloud env? Should we also create a new key?

functions/airtable-connector.js
```diff
@@ -1,6 +1,9 @@
 const Airtable = require('airtable');
 const functions = require('firebase-functions');
+const csv = require('csv-parse');
+const fs = require('fs');
 const zipcodes = require('zipcodes');
+const { geocodeAddress } = require('./geocode.js');
```
nburt, line 6: is this file checked in?

awong-dev: oops. checked in.

functions/airtable-connector.js
awong-dev: moved. Will generate new key later.

PR #886, opened 2020-05-08
Make embed take up more of iframe height

sunnymui:
Now that the list view is shown side by side with the map, the map embed isn't really taking up the full height available in an iframe embed on desktop. 

Bumping up the embed body height to 70vh to fill up more of the height on large screens. This translates to a min iframe height of ~650px before scroll bars are shown on the embed. Takes up most of the iframe height now at more usable large iframe heights.

ftm-before
ftm-after

Note that pushing the iframe height past 750px starts making the white space gap noticeable again, but achieving true responsiveness is gonna more a more in depth change.

### DIFF
--- a/sass/give.scss
+++ b/sass/give.scss
@@ -4,7 +4,7 @@
 @import "filter-styles";
 @import "breakpoints";
 
-$embed-body-height: 50vh;
+$embed-body-height: 70vh;
 $embed-body-height-sm: 35vh;
 
 html {


ngottlieb:
The issue is that they don't all give us that much vertical real estate and we want the footer to show no matter how small our box is... @linderr can you think of an embed example that has the shortest window for our iframe that we can test this on?

linderr:
Hey sorry, late to the game --- try: https://northernontarioppe.ca/pages/initiatives

linderr:
@sunnymui @ngottlieb I don't think I see a huge difference on this collaborator's site? Just tested on Inspector

linderr:
The height of the iframe script we share now is 800px.
I couldn't see a difference like your screen shots on https://northernontarioppe.ca/pages/initiatives @sunnymui  ?

I think we will just have to ask each collaborator change the iframe height to 1000px to achieve the balanced header to map look??

image 
^1000 px iframe height

sunnymui:
Just to make the exact thing I'm talking about more clear:

image

The blue highlighted area is the area of the actual iframe element i.e. the 800px, but you can tell the actual document content in the iframe only takes up some of that space, leaving extra white space below it. 

You'll notice the large whitespace gap to the next element on that page. That's basically where the list view used to be positioned. Technically, I could just put `overflow: hidden` on the iframe element to hide that space, but I'd rather still have it scroll on overflow, just in case the content somehow gets larger than the frame. That way a user can still access the content within. 

At 70vh body height, the minimum height of the iframe element before scrollbars are shown is 650px. On GetUsPPE, I have that iframe element at 700px. If you want to enforce a larger minimum iframe embed height, you can increase vh to match, I just put 70vh to be conservative as probably the smallest you'd want that embedded map and have it still be usable. 

On the linked example for northontarioppe, that one actually has the issue exaggerated even more because it's using the mobile media query height, so it's only 35vh. Probably need to toggle a class when the list view is hidden so it can show a greater height for the map.

linderr:
@ngottlieb @sunnymui I found an even more exaggerated example to test on 😄 
https://coronacourier.nyc/#/resources enjoy

sunnymui:
Lol a 300 px height on that iframe is just asking for a bad time. 

You'd have to have a height media query or window.innerHeight detection that applies a bunch of styles shrinking or hiding elements, significantly changing the layout to make the map halfway usable. Maybe that can be a future upgrade.

Increasing the embed body height to 70vh in this extreme case only makes a slight difference in the visibility of the bottom bar w/ edit map button--both require scrolling down to see the text.

linderr:
Hahah I don’t even know why this is happening.. checked outbound emails, team definitely linked them 800px, not sure what happened on partner’s end - miraculous part is this page drives traffic but it’s probably reverse UX/UI psychology when embed is basically unusable & ppl just come to the main so.. 👌🏻

PR #887, opened 2020-05-08
Add group_type and min_request fields for makers

awong-dev:


client/locations-list-map.js
```diff
@@ -292,10 +292,12 @@ function createMakerMarkerContent(entry, separator) {
 
   addParagraph($.i18n('ftm-makers-website'), entry.website);
   addParagraph($.i18n('ftm-makers-contact'), entry.public_contact);
+  addLine($.i18n('ftm-makers-group-type'), addSpaceAfterComma(entry.capabilities));
```
linderr, line 295: interesting.

linderr:
@awong-dev I don't think you're calling the right column of data for Group Type (function). You are re-displaying Capacity (items)
image

linderr:
Should be this column of data: 
image

client/locations-list-map.js
linderr: I think entry.group_type if I understand correctly?

linderr:
Please wait for Smartling :)

Issue #843, opened 2020-05-05
Expose timestamp in list/map view


4kbt:
As our dataset ages, our listings span more than six weeks of time. While we are working on refreshing as many entries as we can, we would like to let donors know how recently an entry was last updated.  By surfacing this information, we may not need to expire entries off the list as quickly.

A timestamp field is already in data.json. We could expose it to donors as "Last Updated: ".

This is in service of adding a filter to allow donors/makers to select an age range to view.

linderr:
Are you envisioning an actual time/date stamp (which might be on the side of too much information)?
Or "this week!" "last week" "2 weeks ago" etc?

4kbt:
I'd imagined the date/timestamp. It'd be pretty to strip the time and leave
just the date, but printing the existing value in data.json is the minimum
step needed to get the feature added.

nburt:
I'd vote for just a simple date format like MM/DD/YYYY (format localized by locale / country). We can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat for intl formatting.

saikatbhadra:
Just merged, feel free to test it out

PR #899, opened 2020-05-09
Add timestamp to list and map views

saikatbhadra:


client/locations-list-map.js
```diff
@@ -349,6 +350,15 @@ function createRequesterMarkerContent(entry, separator) {
     );
   }
 
+  if (timestamp) {
+    const date = new Date(timestamp);
+    const options = { year: 'numeric', month: 'long', day: 'numeric' };
+    contentTags.push(
+      ce('div', 'label', ctn($.i18n('ftm-date-added'))),
+      ce('div', 'value', ctn(date.toLocaleDateString(undefined, options)))
```
ngottlieb, line 358: is `undefined` here defaulting it to the browser's locale? I feel like since we have a bunch of logic dealing with locale, we should use the calculated locale here to ensure that there aren't weird edge cases

saikatbhadra: Yes, it's defaulting to browser. I don't know how to pull that in, how do I get the user's locale? I need the country and language so it's a full locale like: "de-DE"

saikatbhadra:
This closes: https://github.com/findthemasks/findthemasks/issues/843

Agree that it should be pulling in the calculated locale rather than the browser default before merging. Not sure how to do that though.

client/locations-list-map.js
ngottlieb: country code is accessible as `gCountryCode` -- https://github.com/findthemasks/findthemasks/blob/master/client/locations-list-map.js#L22

I'm not 100% sure how to get `locale` in this module...you could add `getCurrentLocaleParam` to the i18nutils import (`import { getMapsLanguageRegion } from './i18nUtils.js';`) and then just use that function. Surprising that we don't have it already being used in `locations-list-map` anywhere but I guess we've gotten away without it

saikatbhadra: I don't think gCountryCode is correct. I just checked and we are using "uk" for the United Kingdom when it's ISO code is "GB" I think (unless UK works on browsers too)
cc @nburt

saikatbhadra: Ok I think your suggestion works. It seems like we have a UK bug but since we don't have different text for Canadians, Brits, or Americans it doesn't matter.

PR #906, opened 2020-05-10
HACK: Always use the makers-us datafile contents

awong-dev:
All data is in the makers-us.json datafile so serving that file
for all country-specific URLs.

### DIFF
--- a/rootRoutes.js
+++ b/rootRoutes.js
@@ -77,8 +77,9 @@ router.get('/data(-:countryCode)?.json', (req, res) => {
 });
 
 router.get('/makers(-:countryCode)?.json', (req, res) => {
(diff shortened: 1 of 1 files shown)


rootRoutes.js
```diff
@@ -77,8 +77,9 @@ router.get('/data(-:countryCode)?.json', (req, res) => {
 });
 
 router.get('/makers(-:countryCode)?.json', (req, res) => {
-  const countryCode = req.params.countryCode || 'us';
-  sendDataJsonFromCache(cachedMakersData, 'makers', countryCode, res);
+  // All make data is in one file right now. Currently calling it us, though
+  // "global" is probably the right name.
```
linderr, line 81: 😄

PR #907, opened 2020-05-10
Shift out widget creation and use template strings.

### LINKED ISSUE #902: Create templates for the marker and list views
It would nice to use templates rather than JS to render the views inside the list and map views. This would clean the JS code up and also make it easier for less technical users to edit the look and content.

awong-dev:
@saikat and @nick were noting that non-devs have a hard time making
even small content fixes because of the dynamically generated DOM.

They suggested using handlebars in the client but looking at our usage
it seems a template string might sufficient. Prototyping by moving
the widget content generation out into a separate file and replacing
the explicity dom node construction with just an HTML blob that is
given the the browser parser.

Fixes #902


linderr:
Are we still doing this or nah?

awong-dev:
stale. closing.

PR #909, opened 2020-05-11
add i18n for titles in regional spotlight

### LINKED ISSUE #885: Add i18n tags to About Us team positions
Our names can be English but the positions and team titles should be translated at least. Right now all display in English-only.

ngottlieb:
resolves #885 

do not merge until smartling is done


linderr:
Travis is not happy with "auto cancellation" which I have gotten a few times / no idea what it means lol 
The standardized i18n labels are very logical and easy for the Translators to work through, thanks!

ngottlieb:
The Travis failures happen because Smartling submits the PR with a different commit for each language -- so Travis starts running on each commit, then autocancels because the same branch has a new update. It's not a real failure, the only important one is the final travis run

linderr:
Interesting. That makes it so much clearer - thanks!

PR #911, opened 2020-05-11
first pass at click-to-zoom table/map interaction

### LINKED ISSUE #864: Make Contact Card div interactable + center map on pin
Matching map view to list was great.
**However, there is no reciprocal way to control the map through the list.**

After entering their hometown or state into the search bar, a user could scroll down the left list and find a listing they like but has no ability to navigate to the exact listing.

**UX/UI dev solution proposed is to make the entire left-hand-side Contact Card div interactable** and ~~linked to a ?coords= based logic first coded by @ngottlieb:~~ do what @ngottlieb suggests below in comments:
image

**This would also help with frustrations on mobile** where it is unclear when users tap on the Final Address hyperlink they are going to Google Maps instead of staying on the mobile site #837 Ask 1.
(additional UX/UI mobile v2 designs are in-progress to also address this concern)

Full size dev notes on Figma: https://www.figma.com/file/qbibGHgcV1LBcpvd2EoHSM/FindTheMasks?node-id=183%3A5881

ngottlieb:
resolves #864


ngottlieb:
@linderr want to give this a go when the deploy finishes? Note, it also applies to the embed page -- opening a collapsed table entry (i.e. clicking) zooms the map and opens the marker.

linderr:
Yo. This is AWESOME.
+1000 to this working on embed too!

Nice call on the hover colour change.

Definitely will ask the Comms team to take a short video tutorial of this new feature for socials this week!

nburt:
2 questions:

1. Is it possible to keep the map zoomed out a little bit? One thing with opening the marker on click is that it zooms fairly tight and hides most other markers. A little annoying if you're just clicking around for a few different ones and then have to zoom out 
1. Do you think we want to focus on the marker when you open the list item? One thing that I'm wondering is with panned out views if people are just curious and clicking to expand and similar to above, now they're dropped into a very specific location. Do you think it should trigger the focus when clicking on a open list item for embed

linderr:
For 2. I definitely think we should center on the pin when someone clicks on the left-hand contact card to show where it is on a map. As for zoomed in/focused that view is perhaps can use the searchbar zoom logic which I believe shows 3 pins and/or clusters worth of a "view"?? Would that help?

linderr:
Also if someone does click a pin, do we want the left-hand list to also match what they are clicking or no?

ngottlieb:
My initial attempt did what @linderr is describing -- it just zooms to the pin, using our "show at least 3 pins" logic, but does not open the marker text. That might make more sense TBH.

I'm not sure about what click a pin would do in terms of action in the list view besides what already automatically happens when the map moves. We could add the `hover` state to list view cards when a map pin is moused over though.

That's a good point re: embed although getting it to work may be tricky I think...

linderr:
I think that logic does work nicely for rural areas where there are not much pins. Sorry was in calls earlier, going to play around now for urban areas and get @nburt the list of facility categories/colors for Layers. 

image

linderr:
Okay, I can see how @nburt's comment makes sense for well-pinned areas like NYC. The 3-pin logic gets you super zoomed in, especially in some neighbourhoods where the nearest two pins are just down the street. 

Can it be a 3-pins OR zoom=16 logic, whichever is "farther"/less zoomed in?

image

ngottlieb:
I haven't totally resolved any of these things, but I changed the click behavior back to what I'd originally done and added bi-directional mouseover highlighting...maybe both of you give it a try and see what you think now? @nburt @linderr

linderr:
Oh cool - the bidirectional hover is a great idea; I like that we don't have to commit to a click and just scrolldown the left-hand-list and just figure out which pin each card corresponds to! I can't upload a video here so I'll just continue on #dev ?

ngottlieb:
- added analytics events
- made zoom icon that controls zoom action, removed zoom action from full card
- added centering to the zoom behavior 

@linderr @nburt et al

linderr:
This is super cool.

client/locations-list-map.js
```diff
@@ -868,8 +894,9 @@ function createRequesterListItemEl(entry) {
   entry.domElem = ce('div', 'location');
   const header = ce('div', 'd-flex');
   const headerHospitalInfo = ce('div', 'flex-grow-1');
+  const headerZoomLink = ce('div', 'icon icon-search entry-zoom-link');
```
nburt, line 897: Can you add an aria-label to the div? Something like "Zoom to marker"?

ngottlieb: added

nburt: ship it!

Issue #815, opened 2020-05-02
Find the makers nav button not full width on mobile


ngottlieb:
![Screenshot_2020-05-02-15-51-39](https://user-images.githubusercontent.com/1222063/80894118-07f51100-8c8d-11ea-934f-450a8681f918.png)

PR #912, opened 2020-05-11
Fixes findthemasks/findthemasks#815

smattingly:
First PR, and I'm not used to working with forks. If I'm not following protocol, please let me know the right way! Thanks.

### DIFF
--- a/sass/style.scss
+++ b/sass/style.scss
@@ -778,10 +778,19 @@ dl.faq-question {
     padding: 0 1rem 0.5rem 1rem;
   }
 
+  .navbar.ftm .xlink {
+    padding: 0 1rem 0.5rem 1rem;
+  }
+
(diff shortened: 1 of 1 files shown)


ngottlieb:
@smattingly thanks! We typically clone the repo directly and open PRs that way, avoiding the use of forks. It keeps things simple and has the added benefit of triggering our CI builds and Heroku review app builds. 

Are you able to do that? Do you need permissions for that somehow? I haven't onboarded someone recently, can look into it if you're not able to clone the repo as is.

smattingly:
@ngottlieb Yes, that's the workflow I am more accustomed to.  I did not have push rights, but I do now thanks to the invitation from @linderr. Going forward I will push to a feature/fix branch and PR from that. Thanks!

Issue #742, opened 2020-04-28
Add email contact link to marker infowindows


nburt:
Currently they only display in the list of locations

nburt:
Closed by https://github.com/findthemasks/findthemasks/pull/915

PR #915, opened 2020-05-12
Attempt at issue 742

smattingly:
This largely addresses #742 by adding an email link to the map marker info windows. I tested it as far as activating the mail modal, but not beyond that. 

There are a couple of wrinkles that I was unsure how to handle.
1. There appears to be no i18n label for email like the other data in the info window. I'm using the same internationalized string that is used in the main list. This seems to be okay except for missing a colon.
2. When the map is in full screen view, the new link causes the mail modal to appear "under" the map-- that is, it's not visible until the map is toggled back to the default view. I can keep digging on this, but thought someone might already know the fix.


linderr:
Hey @smattingly thanks for diving into the issues upon joining the FTM community!! 
I've suggested @ngottlieb & @nburt to have a look at your PR here this time, but feel free to copy the URL of your PR into our #dev channel on Slack and anyone can chime in or have a look!

Also FYI, Slack appears to be having issues right now :)

client/locations-list-map.js
```diff
@@ -350,6 +351,13 @@ function createRequesterMarkerContent(entry, separator) {
     );
   }
 
+  if (encryptedEmail) {
+    contentTags.push(
+      ce('div', 'label', ctn($.i18n('ftm-email-contact'))),
```
nburt, line 356: Since everything else has the colon can you create a new string?

It goes in `en.json` and you could add something like `"ftm-info-window-email-contact": "Email:"`

From there it'll be google translated automatically and a PR will be opened that you can merge for all the other language translations. Full process is documented here in more detail: https://github.com/findthemasks/findthemasks/wiki/Translation-Process

smattingly: @nburt An update, FYI: I am still working this; the stackoverflow post you suggested is helpful, but leads to other difficulties I'm trying to resolve. Hope to have something today.

Translation should be no problem.

smattingly: @nburt I just pushed [changes](https://github.com/findthemasks/findthemasks/pull/915/commits/649d5e00926d048f858c8e4f7fa5c6d6e3fae004) to fix the contact modal in full screen mode. I started with the stackoverflow solution you shared, which adds a map control so that it will appear above the map in full screen. That worked, but causes Google maps to "take over" the presentation of the contact modal, messing up things after leaving full screen mode.

I spent a long time trying to "take the modal back" from Google maps before moving to the current approach. Since this is heavier front-end work than I typically do, I wanted your feedback on the approach. When entering full screen mode, the code clones the contact modal and adds that clone to the map controls. When leaving full screen mode, the map control and clone are destroyed.

This seems to work well visually in Chrome. I doubt very much that the send functionality will work in full screen mode because that code refers to the pre-existing modal by ID, not the clone. If you approve of this approach, I will address that, and will definitely need to test message sending. I would also modify markup so that some selectors can be class- instead of id-based, in order to avoid duplicate ids when cloning. 

This is a little out of my depth, so if it's not a good approach, please say so. Thanks.

nburt: Just re-deployed the review app and played around with it, this looks great. I realized it'll be tough to test on the review app since the captcha is scoped to the domain. We could test the send from local.findthemasks.com or just do it on prod. I think that once you get the fullscreen submit to work this is good to go!

smattingly: The fullscreen fix works on Chrome and Firefox. Not Edge: it behaves as if fix were not in place: modal is "under" the map, still there if you leave full screen.

nburt:
Merging, awesome work!

PR #923, opened 2020-05-14
add filters to makers for products + capabilities

nburt:


linderr:
Awesome work! Great choice on the two base filter items.

Can the filter options be in alphabetical order?
Categories: 3D printing, A B C, Other
Products: ABC, Other ?

nburt:
> Awesome work! Great choice on the two base filter items.
> 
> Can the filter options be in alphabetical order?
> Categories: 3D printing, A B C, Other
> Product Type: ABC, Other ?

Unfortunately we can't without a substantial amount of work since it'll need to work across all languages. I agree that's better behavior but we should do it in a separate issue since this isn't what we do for find the masks either.

linderr:
You have some mega fan love from Claire C on #makers-ftm from Nation of Makers She also asked for "Other" to be dropped to the bottom --- maybe this is a more appropriate ask; I'll midnight issue that ABC stuff --- thanks, Nate! This is great!

awong-dev:
An actual use of reduce()!!!! <3. 

Very nicely done :)

PR #936, opened 2020-05-16
Add support for getusppe regional affiliates in json.

awong-dev:
Slightly hacky on the functions side, but also not sure it's worth refactoring...

### DIFF
--- a/functions/index.js
+++ b/functions/index.js
@@ -34,7 +34,7 @@ const SHEETS = {
   it: '1YHt6G1ghcXrRqXflevxHAwg2XOXmG2Cym6nSS5vXe7Q',
   pl: '10EsvvozwLTQpn0ejvPjZSWt9lve3fnMHBHltW-v_zkY',
   pt: '1QnyjUUBT_P476dEl0WfQwVnW15Ie7ogty7DiOkMhHLo',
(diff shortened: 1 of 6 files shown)


functions/index.js
```diff
@@ -497,87 +490,25 @@ function toDataByLocation(data) {
...
   const country = get_country_from_path(req);
   if (country === 'makers') {
     loadMakerData(admin, req, res);
     return;
   }
 
-  if (!(country in SHEETS)) {
-    res.status(400).send(`invalid country: ${country} for ${req.path}`);
-    return;
+  let data = null;
```
nburt, line 500: should data = 'data' by default?

nburt:
I'm seeing that the contact name is in the json file, do we want that or should we be filtering out for privacy?

nburt:
I'm seeing that the contact name is in the json file, do we want that or should we be filtering out for privacy?

functions/index.js
awong-dev: I think I prefer it as it is?  Can change later if it becomes a problem.

awong-dev:
@nburt The names are the public contact names so they should be displayed.

Thanks for wiring in the override!  Gonna merge for now so functionality is deployed.

Issue #937, opened 2020-05-17
Narrow screen layout defects


drwestco:
On narrower screens, a couple of layout problems:

"Join Us" link wraps to two lines, screws up vertical alignment of navbar links.

Horizontal scrollbar appears. Scrolling to the right leads to white gaps at navbar and banner.

![image](https://user-images.githubusercontent.com/743030/82156685-8cb96080-9831-11ea-907f-d9380449a4b6.png)

Issue #913, opened 2020-05-11
"Add Makerspace" button styled differently on mobile nav


ngottlieb:
<img width="500" alt="2020-05-11_15-56-59" src="https://user-images.githubusercontent.com/1222063/81620162-1388bc00-93a0-11ea-85ec-37cd1b2425c2.png">

Issue can be seen by switching from localhost:3000 to localhost:3000/makers and looking at the menu with small viewport. 

probably best to change styling of this button to match the Find the Masks button (and the "Add Donation Site" button that displays on the masks dataset)

Issue #873, opened 2020-05-07
Search bar placeholder text gets obstructed on mobile


saikatbhadra:
When the viewport width is very small (like on mobile) the placeholder text "Search by city, zip or address" gets cut off. We should see if we can cut down this text, change its sizing, or do something else to make sure it's still fully visible on all common views.

This bug probably exists in the main page as well

linderr:
“Search by location”? 
![4539B6BB-BBA1-4B08-B74E-73BCB76A231A](https://user-images.githubusercontent.com/62862914/81894798-898a5000-9565-11ea-86b3-8782141d1833.png)

PR #938, opened 2020-05-18
Small fixups

ywyw:
Address some small UI fixups:

#873, #913, #937 

To handle the layout on larger tablets or smaller desktop screens, shrink the font size and spacing a bit like so:

Screen Shot 2020-05-17 at 9 06 30 PM

### DIFF
--- a/sass/style.scss
+++ b/sass/style.scss
@@ -551,6 +551,8 @@ dl.faq-question {
   display: flex;
   min-height: 100%;
   align-items: center;
+  padding-left: 1rem;
+  padding-right: 1rem;
 }
 
 .navbar.ftm .navbar-brand {
--- a/sass/style.scss
+++ b/sass/style.scss
@@ -596,6 +596,10 @@ dl.faq-question {
   }
 }
 
+.navbar.ftm .nav-item {
+  white-space: nowrap;
+}
+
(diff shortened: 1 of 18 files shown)


sass/style.scss
```diff
@@ -940,3 +944,19 @@ div.figure img {
...
   height: 120px;
 }
+
+/****************************************************************
+  * NAVBAR LARGE TABLET / SMALL DESKTOP SIZE TRANSITIONAL STUFF *
+ ****************************************************************/
+
+@media (max-width: 1265px) and (min-width: $breakpoint-md) {
+  .navbar.ftm .navbar-nav .nav-link {
+    font-size: 10.5px;
```
ngottlieb, line 954: can you change this (and line 960) to `rem` to be consistent with our other font-size specs? Also, I think we can get away with it being a bit bigger, at least testing it in Brave right now. I tried `0.8rem` and it worked

ywyw: Done!  I was testing widths all the way down to `992px` and `0.8rem` is a little bit too big for the `iPad Pro` type of screen.  `0.67rem` is about as big as we can go here

ngottlieb: sounds good thanks!

PR #953, opened 2020-05-22
Refresh list once

smattingly:
I noticed that `getFlatFilteredEntries` was being called twice on page load, which seemed kind of expensive. Both calls were coming from `refreshList`, which was called once explicitly, and once because map bounds were "changing" to their initial values.

I thought it might be pretty simple to avoid that redundancy. Much later, I have this, which eliminates all of these unneeded(?) steps:
- the redundancy already mentioned
- building the location list when `?hide-list=true`
- building the map markers when `?hide-map=true`

It also displays stats when `?hide-map=true`, which does not now happen in production. I think because of a bad/outdated jQuery selector that I modified.

### DIFF
--- a/client/locations-list-map.js
+++ b/client/locations-list-map.js
@@ -704,13 +704,13 @@ function updateStats() {
 
   const prettyMarkerCount = numberFormat(countShown, 0);
   const prettyTotalCount = numberFormat(totalEntries, 0);
-  const stats = $('#list-stats');
-  stats.show();
+  const $stats = $('#list-stats');
+  $stats.show();
 
   if (gDataset === 'makers') {
-    stats.html($.i18n('ftm-makers-count', prettyMarkerCount, prettyTotalCount));
+    $stats.html($.i18n('ftm-makers-count', prettyMarkerCount, prettyTotalCount));
   } else {
-    stats.html($.i18n('ftm-requesters-count', prettyMarkerCount, prettyTotalCount));
+    $stats.html($.i18n('ftm-requesters-count', prettyMarkerCount, prettyTotalCount));
   }
 }
 
@@ -1135,10 +1135,6 @@ function initializeEmbedLocationCollapse() {
 }
 
 function refreshList(data, filters) {
-  if (!showList) {
-    return;
-  }
-
   gLocationsListEntries = getFlatFilteredEntries(data, filters);
   gLastLocationRendered = -1;
   $('.locations-list').empty();
--- a/client/locations-list-map.js
+++ b/client/locations-list-map.js
@@ -65,6 +65,8 @@ let gLocationsListEntries = [];
 let gLastLocationRendered = -1;
 
 const searchParams = new FtmUrl(window.location.href).searchparams;
+const showList = searchParams['hide-list'] !== 'true';
+const showMap = searchParams['hide-map'] !== 'true';
 
 // tracks number of entries in dataset
 // gets set when we retrieve the dataset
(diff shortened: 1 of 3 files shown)


client/locations-list-map.js
```diff
@@ -679,20 +681,36 @@ function numberFormat(number, decimalPlaces, decSeparator, thouSeparator) {
...
+
+    // ... but defer to count of markers in map bounds, when applicable.
+    const countInBounds = (count, marker) => count + mapBounds.contains(marker.getPosition());
+    countShown = gPrimaryMarkers.reduce(countInBounds, 0);
+    countShown += gSecondaryMarkers.reduce(countInBounds, 0);
+  }
+
+  const prettyMarkerCount = numberFormat(countShown, 0);
+  const prettyTotalCount = numberFormat(totalEntries, 0);
+  const stats = $('#list-stats');
```
ngottlieb, line 707: it's convention to prefix jQuery nodes with `$` (i.e. `$stats`)

client/locations-list-map.js
```diff
@@ -1130,9 +1135,14 @@ function initializeEmbedLocationCollapse() {
 }
 
 function refreshList(data, filters) {
+  if (!showList) {
```
ngottlieb, line 1138: do we need the `showList` conditional here *and* where the function is called? seems redundant but I might be missing something

client/locations-list-map.js
```diff
@@ -1524,14 +1540,22 @@ function initMap(data, filters) {
...
       const currentLng = mapBounds.getCenter().lng();
 
-      if (currentLat !== gCurrentViewportCenter.lat || currentLng !== gCurrentViewportCenter.lng) {
-        refreshList(data, filters);
+      if (showList) {
+        // When "changing" to initial bounds, the location list is already in sync,
+        //  *unless* the URL specified coordinates.
+        if ((gCurrentViewportCenter.lat && gCurrentViewportCenter.lat !== currentLat)
+          || (gCurrentViewportCenter.lng && gCurrentViewportCenter.lng !== currentLng)
+          || searchParams.coords) {
```
ngottlieb, line 1548: what about if the URL specified a search query `q=`?

ngottlieb: also, can this conditional just be:

    gCurrentViewportCenter.lat !== currentLat || gCurrentViewportCenter.lng !== currentLng || searchParams.coords

seems like if `gCurrentViewportCenter.lat` is undefined or null or otherwise falsey then the condition will be false without having the additional clause?

ngottlieb:
Functionality looks good, my only concern is essentially an unrelated bug that's been made more obvious by this change:

`/embed?hide-map=true` yields a list view that only shows the first 40 entries because infinite scroll isn't set up on the embed. This is now obvious because we have a counter that says, "displaying 3907 requesters."  I suspect no one is using that query param on the embed, but I'm confirming with @linderr on slack, if yes we may need a new issue to solve that.

smattingly:
As discussed on Slack, the scrolling problem is not connected to `?hide-map=true`.

client/locations-list-map.js
smattingly: Will do.

client/locations-list-map.js
smattingly: I debated with myself about that. Decided to call the redundancy "defense in depth". I'll change it.

client/locations-list-map.js
smattingly: No, the first time this runs, `gCurrentViewpointCenter` has its initial value `{}`, so `.lat` is undefined and is `!== currentLat`. That's what triggered the redundant call to `refreshList` that started me on this.

I did try changing `gCurrentViewpointCenter` to be `null` initially, but that caused some other problem that I've forgotten, so I went with this.

views/give.handlebars
```diff
@@ -70,8 +70,8 @@
       </div>
     </div>
 
-    <div class="col-12 {{#unless hideMap}}col-lg-4{{/unless}} order-lg-1 locations-container" style="display: none;">
-      <div id="locations-list {{#if hideMap}}w-100{{/if}}" class="locations-list"></div>
+    <div class="col-12 {{#unless hideMap}}col-lg-4{{/unless}} order-lg-1 locations-container" style="height: 100px; display: none;">
+      <div id="locations-list" class="locations-list {{#if hideMap}}w-100{{/if}}"></div>
```
smattingly, line 74: @ngottlieb I wanted to call attention to this change, since I don't know much handlebars. Looks to me like a Bootstrap css class was sometimes added as a "second id"? When not added, that left a trailing space in the id that was tripping me up.

linderr:
Checked deployment --- infinite scroll is now indeed infinite on embed! Nice catch @smattingly

linderr:
@smattingly tried it out on inspector looks good - sent videos on #dev

awong-dev:
@nburt is out for the weekend.  @ngottlieb do you feel comfortable completing the review? If not, I'll do it tomorrow and merge. This looks like a very nice set of fixes.  Thanks @smattingly !

views/give.handlebars
ngottlieb: Hah. Yeah, 100% a typo/accident. thanks for catching it!

ngottlieb:
new changes look good, thanks @smattingly

PR #956, opened 2020-05-23
Support quick-click "lgtm" and "remove" buttons in mailchimp

### LINKED ISSUE #856: A "Looks Good" link for our refresh-emails
@linderr has a good idea for a way to improve our information-gathering from our refresh-emails. 

At the moment, we ask them to fill in the entire form in order to update us. While this gives us better insight into their changing needs, it takes a lot of time.

What would we need to do in order to surface a link that could make the "LGTM" button make an addition to the Sheet?  It should be straightforward to be able to pass the row-ID to links in the refresh-email, perhaps even as a pre-fill to another Google Form.

Ideally, it would appear as a standard "Edit" entry in Combined, with a correct new timestamp, echoing down everything from the relevant row, and have a little annotation denoting that it was from the LGTM button.

There may be a clever hack that could let us deploy something quickly; I just can't quite see the implementation yet.

awong-dev:
When complete, fixes #856

### DIFF
--- a/apiRoutes.js
+++ b/apiRoutes.js
@@ -1,0 +1,29 @@
+const express = require('express');
+const https = require('https');
+
+const router = express.Router();
+
+router.get('/command/:cmd', (req, res, next) => {
+  res.status(200).send('Done');
+
+  const options = {
+    hostname: 'functions.findthemasks.com',
+    port: 443,
+    path: `/exec/${req.params.cmd}`,
+    method: 'GET',
+  };
+
+  let newData = '';
+  const dataReq = https.request(options, (dataRes) => {
+    dataRes.on('data', (d) => { newData += d; });
+    dataRes.on('end', () => {
+      if (dataRes.statusCode === 200) {
+        console.log(dataRes);
+      } else {
+      }
+        console.error(dataRes);
+    });
+  });
+});
+
+module.exports = router;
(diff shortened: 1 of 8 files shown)


awong-dev:
@nburt yeah, still slowly hacking on the api that receives the URL.  I also changed from gcm to cbc... realized the packet length gets super long with gcm and the damage of a successful cbc attack just isn't that large.

Regarding actions "r" should remove and annotate the "Reason" column with "r"

For "g", that's more tough. Per @4kbt , ideally we want to submit a new form entry (so talk to the form, not the sheet) using the same data as the row. Remove the old row with reason DU.  However, I'm not quite sure why we need this..

4kbt:
The reason is really one of bookkeeping -- a moderator's perception of the
sheet is pseudo-linear in time. In particular, Combined is the only
usable/searchable record of past moderator's decisions. By rolling the
updated entries forward and marking the old entries as "duplicate-updated",
it is possible to understand how the current state of the sheet came to
be.  If we were to simply update the entry itself, moderators have to jump
all over the Sheet to stay on top of updates.

In addition, this pseudo-timeseries approach will also let us do
straightforward analytics on how quickly old entries are getting updated.

If we're able to recopy the row to the bottom without breaking the normal
Form submission, there's no need to go through the Form itself.

awong-dev:
We can definitely just add to the bottom of the sheet without going to the
form, but now the "source tab" relation won't be quite right...

4kbt:
Ah.  True. Going through the Form does make for even more raw-input record
consistency/analytics.

PR #957, opened 2020-05-24
Demo third-party links with real external web site

smattingly:
You can test this at https://dewv.net/ftm/test.html, where I have set up a demo of a mechanism for embedded third party links.

(Note second magnifying glass icon is temporarily used for the third party link.)
I think this approach is simple and reasonably abuse-proof.
For any and only sites that we put in the partner list, their embed requests will be served with links in the embed location list. Those site(s) will just need to handle `/ftm?id=X` requests.

### DIFF
--- a/applicationRoutes.js
+++ b/applicationRoutes.js
@@ -8,6 +8,7 @@ const localizePartialPath = require('./viewHelpers/localizePartialPath');
 const getLocalContactEmail = require('./viewHelpers/getLocalContactEmail');
 const setBananaI18n = require('./middleware/setBananaI18n.js');
 const setCurrentUrl = require('./middleware/setCurrentUrl.js');
+const linkPartners = require('./linkPartners.js');
 
 const herokuVersion = process.env.HEROKU_RELEASE_VERSION;
 
@@ -43,15 +44,14 @@ router.get('/faq', (req, res) => {
   });
 });
 
-const linkPartners = [
-  'https://dewv.net/', // TODO build real site list; this is for testing (Steve Mattingly's site)
-];
-
 router.get(['/give', '/give.html', '/embed'], (req, res) => {
+  // Set up partner links, when applicable.
+  // req.headers.referer = "https://dewv.net/ftm/test.html"
   if (req.headers.referer) {
-    res.locals.partnerSite = linkPartners.find(
-      (element) => req.headers.referer.startsWith(element),
-    );
+    const referer = new URL(req.headers.referer);
+    if (linkPartners[referer.hostname]) {
+      res.locals.partnerSite = `${referer.origin}${linkPartners[referer.hostname]}`;
+    }
   }
 
   const isMaker = res.locals.dataset === 'makers';
--- a/applicationRoutes.js
+++ b/applicationRoutes.js
@@ -43,7 +43,13 @@ router.get('/faq', (req, res) => {
   });
 });
 
+const linkPartners = [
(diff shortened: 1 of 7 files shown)


awong-dev:
Interesting. This looks pretty good!

So basically each specific referrer would need to submit a PR to have themselves added with their own icon.

And then the callback URL format is also similarly encoded (nothing dynamic via parameters).

I like it!

applicationRoutes.js
```diff
@@ -43,7 +43,17 @@ router.get('/faq', (req, res) => {
   });
 });
 
+const linkPartners = [
```
awong-dev, line 46: Can we pull out the various bits where we'd want a partner to change stuff (so the referer allow-list and the icon) into a separate file so it's very easy to know it's 3rd party config？

client/locations-list-map.js
```diff
@@ -1090,6 +1096,9 @@ function getEntryEl(entry) {
     sendEvent('listView', 'clickZoom', entry.name);
     zoomToMarker(entry.marker);
   });
+  $(entry.domElem).find('.entry-partner-link').on('click', () => {
+    window.open(`${document.body.dataset.partnerSite}ftm?id=${entry.row}`, '_blank');
```
awong-dev, line 1100: Is it worth having some config allow for partners to declare their own callback URL?  Or is that oo much for now?

awong-dev:
Tagging @ngottlieb since we were chatting.  This approach seems to meet the security requirements and be even a little bit simpler (ie no URL parameter.)

client/locations-list-map.js
ngottlieb: we could just make the `linkPartners` from `applicationRoutes.js` (abstracted to another file) into an object keyed by the referrer, values being the callback URL?

client/locations-list-map.js
```diff
@@ -952,10 +952,16 @@ function createRequesterListItemEl(entry) {
...
   const header = ce('div', 'd-flex');
   const headerHospitalInfo = ce('div', 'flex-grow-1');
+  const headerOrgType = ce('div', 'flex-grow-1 d-flex justify-content-end text-pink');
   const headerZoomLink = ce('div', 'icon icon-search entry-zoom-link');
   headerZoomLink.setAttribute('aria-label', 'Zoom to marker');
-  const headerOrgType = ce('div', 'flex-grow-1 d-flex justify-content-end text-pink');
-  ac(headerHospitalInfo, ce('h5', null, [ctn(entry.name), headerZoomLink]));
+  const children = [ctn(entry.name), headerZoomLink];
+  if (document.body.dataset.partnerSite) {
+    const headerPartnerLink = ce('div', 'icon icon-search entry-partner-link'); // TODO replace icon
```
ngottlieb, line 960: If we canonically convert `dataset.partnerSite` into a CSS-class-safe string, then all our partners would have to do is add the CSS for their icon (and add themselves to the whitelist in our routing)

applicationRoutes.js
smattingly: I think that's a good idea.

client/locations-list-map.js
smattingly: As @ngottlieb says, it wouldn't be hard to do. However, the original request as it filtered to me sounded maybe like partners were saying "Can you tell us how to ... ?".  You all know the partners and I don't-- your call on whether simple/standard or minor complexity/flexible is better. Or both: allow custom callback but default to `/ftm?id=X` when not specified.

smattingly:
Closing this PR without merge so I can create another to get a redeploy.

PR #960, opened 2020-05-25
first pass at collapsible filters

nburt:
I changed the copy from show/hide to toggle to make the code simpler. I think it might need a little more ux love, especially on the embed page, before merge but putting it our there to get some feedback.


sass/give.scss
```diff
@@ -4,6 +4,14 @@
 @import "filter-styles";
 @import "breakpoints";
 
+.filter-collapse {
+  padding: 0.5rem 1rem;;
```
ngottlieb, line 8: I think it's a little confusing to have padding dependent solely on the order that the CSS is loaded here? I'd say either the CSS here should be more specific (nested under some `embed` specific class) or the `.filter-collapse { padding }` in `filter-styles.scss` should be moved to the main CSS since it's specific to that.

Also, you have two semicolons here.

public/i18n/en.json
```diff
@@ -18,6 +18,7 @@
   "ftm-accepting": "Accepting",
   "ftm-ppe-needed": "PPE needed",
   "ftm-facility-type": "Facility type",
+  "ftm-filter-toggle": "Toggle filter options",
```
ngottlieb, line 21: I feel like it could just say "Additional filters" or "Filters" and that the caret makes it obvious? Just personal preference though.

linderr: I agree the copy should err on the side of ELI5 than assume. Love the choice of "toggle" only concern is "Toggle additional filters" is a longish phrase and worried about mobile width.

What about just colloquial "Toggle more filters" (less characters than current)

Looks good on mobile:
image

linderr:
Embed map definitely needs more UX considerations lol. Will take a video to discuss on #dev b/c github doesn't allow video clip uploads 😞 

Looks good on top referral https://getusppe.org/give/ but most others have a double scroll-bar effect when filters are toggled/opened. Try on https://theppedrive.com/ppe-drive-listing the next top referral after GUP.

GUP has altered their own iframe script to this: 
**<iframe id="give-map" style="width: 100%; border: none; overflow-y: auto; opacity: 1; visibility: visible;" src="https://findthemasks-filters-v3-upehru.herokuapp.com/embed" allow="geolocation"></iframe>**

Let me know if anything should be adopted on our default embed scrip that non-dev outreach members share and troubleshoot with partnerships.

linderr:
Can we make the "X of Y Requests" larger font so it is more prominent? Perhaps almost to the same height as the search bar:
image

May moving the "Reset map & list" option to be right-aligned on the "Toggle filter options" line will look prettier 😃

public/i18n/en.json
nburt: I don't have a strong opinion on it, I choose "filter options" since that was in figma and dropped the "sort" part. If I had to choose I'd probably just go with "Toggle filters" but you tell me

nburt:
> Can we make the "X of Y Requests" larger font so it is more prominent? Perhaps almost to the same height as the search bar: 
> image
> 
> Maybe moving the "Reset map & list" option to be right-aligned on the "Toggle filter options" line will look prettier 😃

I can bump up the font size yeah. That's what I originally had and I changed it cause I think it looks weird with all the spacing not being aligned

image

Maybe if I also right align the requester counts?

image

public/i18n/en.json
linderr: "Toggle filters/Hide Filters" + caret sounds straight forward. Let's do that!

linderr:
Cool let me test out some embeds on mobile view

linderr:
To get around the double scroll bar issue, is the solution to ask external sites to embed at an iframe height of 1000px and then shorten the internal map height to... 750px-ish to accommodate for the height of the filters that may be revealed??? @nburt @ngottlieb

linderr:
toggle looks great on mobile! Way cleaner on the the default load and even open filters don't look clustered. "Reset map & list" would definitely fit on the "Toggle filter options" line on mobile
image

nburt:
> To get around the double scroll bar issue, is the solution to ask external sites to embed at an iframe height of 1000px and then shorten the internal map height to... 750px-ish to accommodate for the height of the filters that may be revealed??? @nburt @ngottlieb

just pushed a few more commits, try the new version once it deploys. I positioned the filters dropdown so that it doesn't push down content but just overlays on top of it. I think that should solve the double scrollbar issues

linderr:
Mobile looks awesome:
image

Desktop: what if we must made the Search bar longer to balance the empty space?
image

linderr:
Awesome yes, no more double scroll bars!

nburt:
Updated again, any thing else?

linderr:
Looks awesome! Thank you!!

4kbt:
These look awesome. Thank you!

PR #968, opened 2020-05-26
Close #966

### LINKED ISSUE #966: Remove ?hide-map=true param
It doesn't fully function right now, it's unused, and no one can imagine a use case for it. So let's get rid of it!

smattingly:


client/locations-list-map.js
```diff
@@ -1701,10 +1696,8 @@ $(() => {
 
     updateFilters(filters);
 
-    if (showMap) {
-      $map.show();
-      loadMapScript(data, filters);
-    }
+    $map.show();
```
nburt, line 1699: is this set to `display: none` by default? By removing this query param do we need that anymore?

smattingly: Good call. Inspired by that, I reworked handling of initial show vs. hide of several elements.

client/locations-list-map.js
```diff
@@ -1579,53 +1576,54 @@ function initMap(data, filters) {
...
-      const elem = document.documentElement;
-      if (elem.requestFullscreen) {
-        elem.requestFullscreen();
-      } else if (elem.mozRequestFullScreen) { /* Firefox */
-        elem.mozRequestFullScreen();
-      } else if (elem.webkitRequestFullscreen) { /* Chrome, Safari and Opera */
-        elem.webkitRequestFullscreen();
-      } else if (elem.msRequestFullscreen) { /* IE/Edge */
-        elem.msRequestFullscreen();
+  if (!isEmbed) {
```
smattingly, line 1579: @nburt The embed page does not show the map full screen control. Has this always been true, or did I remove it when I implemented the custom control? (Early on, I was just working with the home page and not really aware of the embed.)

nburt: hmm I hadn't noticed that, I'm guessing it was left out when you did the custom embed. Any idea why it isn't working there?

views/index.handlebars
```diff
@@ -89,7 +89,7 @@
           {{/if}}
         </div>
         <div id="map"></div>
-        <button id="customFullscreenButton" draggable="false" title="Toggle fullscreen view" type="button" style="display: none; background: none rgb(255, 255, 255); border: 0px; margin: 10px; padding: 0px; position: absolute; cursor: pointer; user-select: none; border-radius: 2px; height: 40px; width: 40px; box-shadow: rgba(0, 0, 0, 0.3) 0px 1px 4px -1px; overflow: hidden; top: 0px; right: 0px;">
+        <button id="customFullscreenButton" draggable="false" title="Toggle fullscreen view" type="button" style="background: none rgb(255, 255, 255); border: 0px; margin: 10px; padding: 0px; position: absolute; cursor: pointer; user-select: none; border-radius: 2px; height: 40px; width: 40px; box-shadow: rgba(0, 0, 0, 0.3) 0px 1px 4px -1px; overflow: hidden; top: 0px; right: 0px;">
```
nburt, line 92: does this just need to be added to give.handlebars to fix the embed fullscreen?

client/locations-list-map.js
ngottlieb: it's quite plausible I lost it when I did the UX2 redesign of the embed... whoops! sorry!

views/index.handlebars
smattingly: Yes, I think so. I just wasn't sure if the fullscreen was wanted on the embed.

Issue #959, opened 2020-05-25
Layer System Legend


linderr:
Increasing number of groups have various datasets (needs, makers, testing centres, case number, risk number, open/shelter-in-place policy status) who want to partner via a data "overlay/layer" system. Users shouldn't have to go to 6 websites for 6 datasets.

**Let's try this logic by just layering on our own Find the Maker data set onto the Find the Masks map, and vice versa.**

@awong-dev suggests: 
- on the Masks map, PPE needs should stay the primary dataset while the Makers is the secondary dataset
- on the Makers map, makers stay the primary dataset while PPE needs is the secondary dataset

Other notes:
- the X of Y counter should not include all datasets (only counts primary dataset Requests v. Groups)
- @ngottlieb thinks secondary datasets do not appear in left-hand list view
- @ngottlieb thinks secondary datasets only appear in map-view and you can click on them to trigger the pop-up info card
- clustering of second dataset should be uniformly one color #d9d9d9 grey
- upon zooming in, single pin resolution will show the following below

@sumitsekhri has suggested to keep the pin outline but use different iconography within the pin.
![image](https://user-images.githubusercontent.com/62862914/82782578-7bfc8200-9e11-11ea-9c68-4189131b9e3a.png)

Makers dataset using **gear icon** in pin. Resources: 
- https://materialdesignicons.com/icon/cog
- https://materialdesignicons.com/icon/cog-outline
- Correct yellow should be Find the Makers yellow #FDC276 (not pictured above)

Reference: sister FTM-FR site shows both needs and makers on one map.
![image](https://user-images.githubusercontent.com/62862914/82783343-1ad5ae00-9e13-11ea-8a86-9bdacd71de5f.png)
https://protegeons-nos-soignants.fr/carte-globale/

nburt:
@linderr a few questions:

1. Do we have the icon with the gear in it yet? What should the "hover" color be for makers? Do we assume that each dataset will have its own icon? (requester = regular, maker = gear, etc.)
2. Sounds like primary and secondary datasets should be clustered separately? Are they clustered as primary vs. secondary or is each dataset clustered separately?
3. @ngottlieb curious about the list-view. Why do you think secondary datasets should be excluded from that? Would it be confusing if you added a new dataset and the X of Y numbers don't change while the map does?
[...]

PR #969, opened 2020-05-27
Layers

nburt:
WIP for https://github.com/findthemasks/findthemasks/issues/959


public/images/markers/requester_marker_hover.svg
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="31pt" height="31pt" viewBox="0 0 31 31" version="1.1">
+<g id="surface1">
+<path style=" stroke:none;fill-rule:nonzero;fill:rgb(76.862745%,21.960784%,18.039216%);fill-opacity:1;" d="M 15.5 2.582031 C 13.101562 2.582031 10.800781 3.535156 9.105469 5.230469 C 7.410156 6.925781 6.457031 9.226562 6.457031 11.625 C 6.457031 18.40625 15.5 28.417969 15.5 28.417969 C 15.5 28.417969 24.542969 18.40625 24.542969 11.625 C 24.542969 9.226562 23.589844 6.925781 21.894531 5.230469 C 20.199219 3.535156 17.898438 2.582031 15.5 2.582031 Z M 15.5 2.582031 "/>
```
linderr, line 4: Whoa.

linderr:
Just had another look - This is soo close to merging! Honestly, I think the legend as you've coded it is fine since it's only Makers +/- Masks datasets right now and logic works great. The design of the legend is not a block and can be revisited when we add other datasets :)

linderr:
Triggering the secondary layer on both Masks and Maker maps on/off at different zooms --- I noticed that due to the densities of PPE need requests and Maker Groups, they tend to cluster in the red and yellow ranges, respectively --- which gives it a vibe that red pins/clusters = needs & yellow pins/clusters = makers in a way that feels intuitive. 

If that's cool with you guys to make all needs clusters red and all makers clusters yellow, regardless of size, we can table second datasets being uniformly one color (#d9d9d9 grey) until we have to deal with more than 2 datasets?

linderr:
@ngottlieb can probably chime whether this would be sacrilegious in map cluster conventions as most maps look like they use different colour clusters to denote size of clusters lol

ngottlieb:
re: clustering, I agree with @linderr that it's a little bit confusing as of now but I don't think it's necessarily worth holding the release up. Ideally we could use different cluster icons for each dataset down the road...

nburt:
I updated the legend with the box-shadow.

Maybe we can do different shades of a single color for each dataset? requesters could be blue, makers orange, etc. I think light -> dark to denote higher concentration?

I don't think that'd be too hard to add. Do we think we need to hold off merge until that or should I release this and then add a follow up issue?

linderr:
The gradient idea sounds great! Let's just make that a separate issue so we can merge this one :)

Let's make it easy for ourselves and just match pin colour to the colour family of the clusters. 
needs = red pins and gradient of red cluster to denote density
makers = yellow/orangey pins and gradient of yellow/orangey clusters

Then as we add more datasets the convention can just be pin colour = cluster colour family instead of dreaming up colour combos every dataset :) 

I'll look for a hexcode resource where we can standardize the gradient to get ready! Excited!! Thanks for doing this @nburt !! 🎉

linderr:
This this the 600th closed PR!!! 🎉 
Thanks for doing this, looks great!

PR #970, opened 2020-05-27
Add ?id= param that zooms to requester marker

smattingly:


client/locations-list-map.js
```diff
@@ -1701,6 +1701,21 @@ $(() => {
...
+    if (id) {
+      const headers = result.values[1];
+      const rowIdIndex = headers.indexOf('row_id');
+      const latIndex = headers.indexOf('lat');
+      const lngIndex = headers.indexOf('lng');
+      const entry = result.values.find((element) => element[rowIdIndex] === id);
+      if (entry) {
+        searchParams.coords = `${entry[latIndex]},${entry[lngIndex]}`;
+      } else {
+        console.warn(`Found no row with id ${id}.`);
```
nburt, line 1715: @linderr @4kbt Do we have a "row_id" column for any country besides the US? If not can we add? It's been a while since I've added a column but I don't think it'd impact anything

client/locations-list-map.js
```diff
@@ -1701,6 +1701,21 @@ $(() => {
...
+    // Translate ?id= param to ?coords= param.
+    const { id } = searchParams;
+    if (id) {
+      const headers = result.values[1];
+      const rowIdIndex = headers.indexOf('row_id');
+      const latIndex = headers.indexOf('lat');
+      const lngIndex = headers.indexOf('lng');
+      const entry = result.values.find((element) => element[rowIdIndex] === id);
+      if (entry) {
+        searchParams.coords = `${entry[latIndex]},${entry[lngIndex]}`;
```
nburt, line 1713: I noticed that for some ids it isn't zoomed in enough, e.g. https://findthemasks-zoommarker-hd7grj.herokuapp.com/?id=5021 displays 93 total markers.

For that one specifically adding zoom=15 solves this, do you think we should just add that in for each "id" param? There might be a better way using the maps api or we could try and click on the marker / highlight it to distinguish but I think we'll also need some zoom since the above request is far enough out that there are some clusters too

smattingly: For some reason, it's not the same result as [manually using the coordinates](https://findthemasks-zoommarker-hd7grj.herokuapp.com/?coords=33.5940329,-117.726433). That's better, but still not good.

You all have good `zoomToMarker` logic, I saw no easy way to get the marker to zoom to. Suppose I can search the collections for it. I'm a little fuzzy on the primary, secondary, other marker collections.

nburt: weird... what do you think about shipping as is? We still do have the zoom param so that partners could add that on top of the id if they feel like it isn't zooming to their liking

smattingly: I've already taken a different approach to use `zoomToMarker`. It's functional and I'm working out kinks (like why isn't the marker centered).

smattingly: It's now working with `zoomToMarker`. At the moment, the locations list is built *3 times* on page load of `?id=`  URLs, but I believe it might be easier to fix that when these changes meet up with my changes in PR #968.

client/locations-list-map.js
linderr: I don't think so - I can go through all non-US datasets and add tonight

client/locations-list-map.js
linderr: @smattingly does location list loading 3 times mean that /reloadsheetdata is being called 3 times? We have seen a spike in map API $ coming from /reloadsheetdata/ the last week and have been trying to figure out why! @nburt

smattingly: Not unless there's some linkage I'm unaware of, @linderr. I first touched this maybe 36 hours ago.

nburt: no we wouldn't be calling reloadsheetdata from anywhere here, it's only called manually or by sheets trigger every 10 minutes

PR #978, opened 2020-05-28
Investigating final link on Who? Makers FAQ

linderr:


ngottlieb:
@linderr this doesn't quite resolve the problem, the problem is in `en.json` -- the text in the template only ever comes through if `en.json` doesn't have a matching key (or if our i18n javascript breaks).

the issue is at the end of that line: `They are also collating <a href=\"https://www.nationofmakers.us/covid-19-maker-response\"resources for makers responding to COVID-19</a>."` doesn't actually close the opening <a> tag. So you just need to add `>` after the URL. https://github.com/findthemasks/findthemasks/blob/master/public/i18n/en.json#L255

linderr:
oi - I skimmed that line like 3 times and didn't catch it!!
A volunteer saw it on our main site so apparently it's been like this for weeks lol!

linderr:
waiting for deployment for final check then merging - thank you!

PR #982, opened 2020-05-28
add zoom icon tooltip

### LINKED ISSUE #980: Add "tool tip" hover to Zoom To icon
@ngottlieb started us off by coded the "Zoom to" function to allow navigation from the left-hand side list to locate an exact listing on the map view.

As we add more functionality like @smattingly's ?id= param or "shopify" features on embed, we will likely add more icons next to Nick's magnifying glass and people need a primer on what each one does.

Let's add "tool tips" when people over on icons in this style:
image
- magnifying icon text (masks): "zoom to request"
- magnifying icon text (makers): "zoom to maker"
- bg colour is #767171

a resource that supposedly adapts to any size view port (mobile-friendly): https://osvaldas.info/elegant-css-and-jquery-tooltip-responsive-mobile-friendly

ngottlieb:
resolves #980 

might want to broaden the code a little bit in the next iteration when we add more icons with tooltips. But the Bootstrap tooltip / click event weirdness makes that harder / more annoying.


linderr:
Looks great!
Is it possible to include if /makers "zoom to maker" and if / "zoom to request" for internationalization?

I see here it's "zoom to marker" on both websites --- ideally text also goes through Smartling so if i'm hovering in another language it will display the tooltip in other languages.

Marker might be translated to a fat pen, or a person who grades, or a marker as you intended :)

ngottlieb:
I was just matching the the `aria-label` (accessibility) terminology we used. But I see your point, and also realize it should i18n and doesn't right now so I'll do that.

ngottlieb:
@linderr (cc: @nburt) added i18n for requesters/makers plus a default i18n string for "other" datasets. Just waiting for smartling.

linderr:
You're not an idiot!

linderr:
Latest deployment looks great. The tooltip you chose looks super cute!! --- also checked on embed and I think something like this will really begin to shift the choice between map-only and map+list view for the embed maps moving forward!! Whenever you feel good about it please merge and i'll train the outreach folks on this new narrative 👍 thank you!!

Issue #986, opened 2020-05-29
Fix legacy browser issues with IE11


linderr:
**FTM does not load on Internet Explore (white page + infinite wheel).**
Many hospitals and government (and labs) still use various non-Windows 10 Windows operating systems and default to IE.

**Here is the errors from the dev console:**
![image](https://user-images.githubusercontent.com/62862914/83292252-a007ec80-a19e-11ea-8d44-0a9bd440c6b5.png)

@awong-dev states the Syntax error is the issue
@ngottlieb muses this can be fixed with babel. We have `.babelrc` so that is a place to start

smattingly:
FYI, the referenced column 38675 is the first appearance in that file of `class` *as a reserved word*.

My babel knowledge is minimal, but I see that we are on version 7.x with a .babelrc, and wonder if [this information](https://babeljs.io/docs/en/config-files#6x-vs-7x-babelrc-loading) is relevant?  Tagging @ngottlieb since he was mentioned in connection with babel.

linderr:
that's interesting, @smattingly I'm going to copy your comment into #dev as well for more visibility :)

smattingly:
I downloaded IE 11 and confirmed that the code fetched from the production site contains class declarations. I assume that means that the babel transpile isn't happening.

smattingly:
Progress! See PR #991.

PR #991, opened 2020-05-31
Site loads in IE 11, but problems remain

smattingly:
Progress on #986. I got the site to load in IE 11 by bringing babel settings into the webpack config file.

However, there are still problems. Solved some by including babel plugin for destructuring assignments.

The current problem: `location-list-map.js:1291` is getting `undefined` back from the `htmlToElements` call.

This seems to be related to the "mobius selectr"? It appears that client code is pulling that in from `node_modules`-- a concept that is foreign to me. I tried getting babel to transpile it, but may not have done it correctly, or I may be barking up the wrong tree.


awong-dev:
@smattingly 
Nice!!! And good find on the configuration style changes.  I'm a little confused at why the .babelrc isn't loading. It makes me think there must be a way to bridge the two files...but if not, we should delete the .bablerc so people don't get confused by it.  Alternately, maybe switch to babel.config.json?  IMO, there should be one single babel setup for the whole project.

So node_modules is basically the equivalent of /lib for projects managed by npm.  When you do "npm install" it drags down everything in package.json into that directory.

Webpack then starts at your entry points and walks the tree of require/imports to all files, many of which are in node_modules, and produces a single packed javascript bundle with a source-map for debugging.

I think we actually _do_ want to transpile everything in node_modules. If we don't then any library code that uses newer JS features will cause legacy browser to break as well.

linderr:
Wow, Steve! Dragging us across this hurdle lol.

Just tried in IE11. It loads but many problems remain to be resolved as you said! Here is a running tally of what I am noticing:
- clusters appear but as you zoom in, they resolve to nothing. There are no individual pins
- similarly @ngottlieb's "zoom to" function will zoom to nothing now :)
- the pins "exist" / are interact-able but are ghost pins: 
image
- @nburt's filter system toggles open but no filter menus show
- top nav is not locked, leaving a nav-height white space above the map view port
- top-left logo is microscopic:
image
- buttons have no top-padding?
image

But other than that, looks great!!!! hahaha <3

linderr:
Going into inspect element we get new errors than before for IE11:
image

linderr:
If I try to view other versions (IE10, IE9 etc) on dev tools, the page is no longer loadable even on IE11 settings and a page refresh results in this pop-up on page load:
image

linderr:
Happy Debugging, guys!!! 🤗 🙏 🎉

smattingly:
@awong-dev I had tried babel.config.json without success, but apparently other pieces were still missing, because I have it working now; backed babel config out of webpack config again.

Re: node_modules, I meant that to me it has always been server-side code; I don't do sophisticated front-end stuff and I had never seen node_modules being pulled into client side code. I tried removing the the node_modules exclude from webpack/babel-loader, but it seemed to make no difference. So I left the exclude in place; apparently the things used in the front end get transpiled anyway because of the dependency arrangements.

smattingly:
Latest commit takes care of the problem related to setup of filter controls with mobius1-selectr. IE11 was not giving the same results from the `createContextualFragment` approach used by `htmlToElements`, so I did it a different way.

smattingly:
The map markers are not appearing because the height and width on the `` tags are zero in IE11. I have tried several ways of overriding that without success. As I understand it, those tags are generated by Google Maps code a couple of layers down from anything we touch directly.

I notice that the clusters, which do appear, have CSS classes like `cluster` and `cluster-2`, which also seem to be applied by GM code. I wonder if there are some CSS classes that we should apply to some of the marker stuff?

smattingly:
So, the marker problem has to do with GMaps and SVG markers in IE11. [This post](https://stackoverflow.com/questions/19719574/google-maps-svg-image-marker-icons-not-showing-in-ie11) and others it links too have details.

Following those, I have badly degraded markers appearing. I will try some more tomorrow, but it may be worth thinking about something other than SVG (the clusters are PNG).

linderr:
Hi Steve, thanks so much for further investigating --- saving to PNG sound good --- The Partner page is also mostly .png files and I noticed the .svg files are displaying incorrectly (look at Slack's logo)

linderr:
Latest deployment has come a long way @smattingly ! I changed the above bullet points to checklist to help keep track. This is really coming along!

smattingly:
@linderr I spent much of the day on the remaining visual/layout issues, without any real progress. (I did manage to size up the FTM logo, but could not position it correctly.)

BTW, there are problems with the full screen map view, too.

I think the root cause of many or all of these problems is that IE11 has a very buggy implementation of flexbox, and no support at all for `position: sticky;` The Bootstrap classes used to style the FTM site's UI rely fairly heavily on those features.

IOW, solving the problems probably requires some extensive re-work of the UI. My skills are not well-suited to that, and I think my time would be better invested on other tasks.

linderr:
Thank you for getting us this far, @smattingly !! It's already a huge improvement over the page not loading at all!!

I found this write up which attempts to address the issue of some browsers not addressing 'sticky' 
https://css-tricks.com/stickybits-alternative-position-sticky-polyfills/

linderr:
Do you want to write an update post on #dev to notify what you've done so far and if anyone is interesting in picking it up?

A work around to the logo is just to omit the logo completely if the page is accessed on IE? The text-based "Find the Masks" banner is already click-able. I think if that were possible, we can merge as is and work on the remainder issues on an iteration.

linderr:
Which page are you referring to for the "full screen map view" @smattingly ? give.html?

smattingly:
@linderr Looks like you got there ahead of me on the #dev update.

I was referring to the Google Maps full screen button that appears on the home page.

I had seen some of those polyfills, but weren't sure if we wanted to go that route. I can take a quick look to see what they might get us.

linderr:
haha actually it was @awong-dev, @smattingly :) we're all very trigger happy here and @nburt came to the review and merge rescue!

PR #1001, opened 2020-06-03
Sticky

### LINKED ISSUE #986: Fix legacy browser issues with IE11

smattingly:
Fixes #986. Resolves all known issues with IE11 presentation.


smattingly:
@linderr check it out. This is twice now-- apparently, the best way for me to learn something is to tell FTM folks I can't do it.

For the logo, I tweaked the svg itself, since I couldn't fix it with CSS.

For sticky top nav, I tried two polyfills without success before just doing it myself.

Hid map full screen from IE11 since it is problematic.

Although the results are good, the techniques, with browser-specific CSS, is justified only by the difficulty/necessity of supporting IE. Future content additions/changes may need to be tweaked to look right in IE.

Front-end code, now `-ms-transform`. I feel as though I have turned to the dark side.

linderr:
Wow @smattingly I just looked up the CSS -ms-transform property and what a great find!
apparently img#upsideDown {-ms-transform: scale(2) rotate(180);} is possible O_O very interesting.

Tested on IE11 and it looks GREAT! I am very sorry for your existential crisis and wish I understood more development to understand why this is you turning to the dark side - but you will always be a beacon of IE11 light for me!! 😄 

Once this is merged, I will be very happy to write to all the "sorry, this doesn't work in IE" emails sitting in our inbox to update them that it does!! Thank you for learning something new and allowing legacy browser support

sass/style.scss
```diff
@@ -55,6 +55,18 @@ a {
   color: white;
 }
 
+.ie-fix-vertical, div.col-8, div.col-10, div.col-12, div.maketitle {
```
nburt, line 58: I don't have IE but looks good to me in Brave. My one question is whether this col modification does anything for other places that use these classes?

awong-dev:
I also don't have IE to test, but this looks good to me! Am holding off on merging since nate had a question, but I'm happy with this fix... vendor specific tags and all. ;)

I wonder if the "proper" fix is to use Autoprefixer in our webpack config: https://github.com/postcss/autoprefixer.  That way you only have to invoke something that's part of the dark side and not actually touch the ms-specific vendor extensions yourself.

all problems in CS (including evil taint) can be solved by an extra level of indirection.... maybe? :P

awong-dev:
we should merge as is though and mess with autoprefix later IMO.

smattingly:
Let me do some more research on this. It appears to be messing up the new alert in Edge.

sass/style.scss
smattingly: @nburt I believe it was okay, but it's now moot-- that css rule is gone.

smattingly:
This has been reworked to use CSS queries to determine when we are in IE10+ *and not Edge*. That distinction was not made before, so Edge was being hobbled. 

In IE, special CSS rules fake the sticky navbar and tweak some other positioning. I didn't want to, but saw no reasonable way around adding one div to top_nav simply to adjust layout in IE.

I don't think we want to use autoprefixer, which I will explain in #dev.

linderr:
Thank you for including the not Edge logic with ms-information!

awong-dev:
Merging!!!!

PR #1002, opened 2020-06-03
Add partner callback links to map infowindows

smattingly:
This can be tested from https://dewv.net/ftm/test.html which I temporarily updated to embed the heroku deployment.

### DIFF
--- a/client/locations-list-map.js
+++ b/client/locations-list-map.js
@@ -382,8 +382,15 @@ function createRequesterMarkerContent(entry, separator) {
     website,
   } = entry;
 
+  const header = separator ? ce('h5', 'separator', ctn(name)) : ce('h5', null, ctn(name));
+  const headerPartnerLink = createPartnerLinkIcon(entry.row_id);
+
+  if (headerPartnerLink) {
+    ac(header, headerPartnerLink);
+  }
+
   // Text to go into InfoWindow
-  const contentTags = separator ? [ce('h5', 'separator', ctn(name))] : [ce('h5', null, ctn(name))];
+  const contentTags = [header];
 
   if (orgType && orgType.length) {
     contentTags.push(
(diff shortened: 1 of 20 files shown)


client/locations-list-map.js
```diff
@@ -382,8 +382,15 @@ function createRequesterMarkerContent(entry, separator) {
...
+  const header = separator ? ce('h5', 'separator', ctn(name)) : ce('h5', null, ctn(name));
+  const headerPartnerLink = createPartnerLinkIcon(entry.row_id);
+
+  if (headerPartnerLink) {
+    ac(header, headerPartnerLink);
+  }
+
   // Text to go into InfoWindow
-  const contentTags = separator ? [ce('h5', 'separator', ctn(name))] : [ce('h5', null, ctn(name))];
+  const contentTags = [header];
```
ngottlieb, line 393: we have a bit of overlap in this area, but it'll be easy to resolve and I can do it on my branch.

smattingly: The links are the "D&E" images that match the site favicon.

ngottlieb: Yeah, it's not that I don't see the links, it's that they're not there. And the `body` tag doesn't have the partner data attributes set...

smattingly: Do you mean that you don't see the red-circled icon/link at the test URL shown?

image

smattingly: Yes, I think that is what you mean. I went to a different machine and don't see them either.

ngottlieb: Yes, I mean exactly that.

I also mean (when I investigated the issue) that the variables your code is setting in the router aren't being set in my browser. I'm using Brave, it's a Chromium based browser.

ngottlieb: I also don't see them in Firefox. They're there in Safari.

ngottlieb: Hmm, this must be some kind of caching issue, but I can't get it to resolve in Brave...

I just opened Chrome (a browser I don't use on this computer) and it worked. So I guess don't worry about it, some sort of local quirk.

ngottlieb:
@smattingly I *think* this is a real issue in Brave, but it may be unique to Brave? I asked @nburt if he can replicate since I think he uses Brave.

ngottlieb:
So @nburt is seeing the same thing in Brave.

I'm looking into the browser spec here and I'm not sure that we can use referrer safely. Behavior seems to be inconsistent, particularly with HTTPS, and it also seems like browsers are not supposed to transmit the original website's URL as a referrer header within an iframe in https though some obviously do.

Maybe we can try to tweak this by using the `referrerpolicy` attribute? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy

smattingly:
@ngottlieb No, it seems the only place it works are the browser*s* on my dev machine.

ngottlieb:
@smattingly not quite, it works on some of my browsers. I think it depends on the browser implementation of `referrer` headers within iframes.

smattingly:
The referrer policy didn't seem to be help, so I went back to the earlier idea of having the link partner identify themselves with a query parameter.

Issue #979, opened 2020-05-28
Enable public use of ?id= param


linderr:
@smattingly has coded the ?id= param to allow unique URLs per listing for outreach & comms teams to share, which is awesome. 

**Let's support this for public use on our main website and the embed map.** 
Example: https://findthemasks.com/ca/?id=67

**[Left-hand contact card:]**
- add  "file-link" icon next to the "zoom to" magnifying glass **and upon click adds URL to clipboard**, if possible
- use icon: https://materialdesignicons.com/icon/file-link
- colour icon dark grey #767171
- tooltip text should be "copy link"
![image](https://user-images.githubusercontent.com/62862914/83127404-f6860580-a08e-11ea-86fc-d298041d381b.png)

**[Pop-up contact card:]**
- add "link" icon next to organization name **upon click adds URL to clipboard**
- use icon: https://materialdesignicons.com/icon/link
- colour icon medium grey #c6c6c6 
- tooltip text should be "copy link"
![image](https://user-images.githubusercontent.com/62862914/83127372-ee2dca80-a08e-11ea-87d2-dc9418308a3a.png)

linderr:
tool tip text enabled by #980

ngottlieb:
@smattingly are you doing this one? I was about to take it then saw you were assigned to it.

smattingly:
@ngottlieb From informal talks with @linderr, I've considered it on my plate for a while. My first step today was to work on a similar issue not actually captured here: putting the partner callback links in the map infoWindow as well as the locations list. I'm still working on that, so I haven't actually touched this work yet. Your call-- take it if you want it.

ngottlieb:
Sure I'll do it, I did the other tooltip thing so was in this area of the code recently

PR #1008, opened 2020-06-03
add row id to makers json airtable connector

ngottlieb:
I actually have not touched the `functions` part of the code before, nor looked at the Airtable, but @linderr tells me `row_id` is in Airtable and it looked like I just needed to add it here to get it in the data-json

required for links to be usable on makers dataset for #979


nburt:
I also assume this would work but I haven't touched anything airtable so would wait on merge / deploy for @awong-dev

awong-dev:
deployed too!

Issue #868, opened 2020-05-07
Add country partial for UK About Us page


linderr:
The UK team has grown and is an active part of the FTM community but currently they are on the v1 all-Europe About Us https://findthemasks.com/uk/whoweare

**They should have their own dedicated About Us page using the regional_spotlight.handlebars logic** 

Updated Content for UK team's Regional Spotlight: 
https://docs.google.com/document/d/1b9lj2j3vqJrZgqfozbZ2reSO6xnC_a00VtKO-xxUB7g/edit#bookmark=id.a51x07aweoz

zhuochenwang:
Can the bears interns try to resolve this issue?

linderr:
I could not tag Steph I believe she still has to accept the contributor invite + do 2FA.

stephmarisa:
Hi Linda! Sorry I just set up 2FA and am all ready now!

linderr:
Closed by #1013 !! Good job!!

PR #1013, opened 2020-06-05
ftm-intern-bears first pull request! (Issue #868)

stephmarisa:
Added country partial for UK and modified en.json file.


zhuochenwang:
We will try to push our code through a branch instead a fork!

PR #1034, opened 2020-06-07
850/rest filter boxes on reset map

zhuochenwang:
I edited the location-list-map.js so we can reference the Selector objects for the selected filters any time we select/deselect them. This allows us to go through these objects and deselect them upon resetting the map.

### DIFF
--- a/client/locations-list-map.js
+++ b/client/locations-list-map.js
@@ -1257,7 +1257,7 @@ function onFilterChange(data, prefix, idx, selected, filters, selectr) {
   }
   const temp = filters[prefix][Object.keys(filters[prefix])[idx]].name;
   if (selected) {
-    filterSelectors[temp]=selectr;
+    filterSelectors[temp] = selectr;
   } else {
     delete filterSelectors[temp];
   }
@@ -1536,8 +1536,8 @@ function initMapSearch(data, filters) {
     e.preventDefault();
     resetMap(data, filters);
     $search.val('');
-    Object.values(filterSelectors).forEach((current)=>{
-      current.clear();
+    Object.keys(filterSelectors).map((current) => {
+      filterSelectors[current].clear();
     });
     sendEvent('map', 'reset', 'default-location');
   });
--- a/client/locations-list-map.js
+++ b/client/locations-list-map.js
@@ -1536,7 +1536,7 @@ function initMapSearch(data, filters) {
     e.preventDefault();
     resetMap(data, filters);
     $search.val('');
-    Object.keys(filterSelectors).map((current) => {
+    Object.keys(filterSelectors).forEach((current) => {
       filterSelectors[current].clear();
     });
     sendEvent('map', 'reset', 'default-location');
--- a/client/locations-list-map.js
+++ b/client/locations-list-map.js
@@ -49,9 +49,6 @@ let gSecondaryCluster = null;
 
 let gCurrentViewportCenter = {};
 
-// New object that stores key:value pairs between a filter and its corresponding Selector object
-const filterSelectors = {};
-
 const gDatasetMarkers = {
   requester: {
     standard: '/images/markers/requester_marker.svg',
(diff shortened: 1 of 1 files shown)


client/locations-list-map.js
```diff
@@ -49,6 +49,9 @@ let gSecondaryCluster = null;
 
 let gCurrentViewportCenter = {};
 
+// New object that stores key:value pairs between a filter and its corresponding Selector object
+const filterSelectors = {};
```
awong-dev, line 53: hmm... I'm not loving more global variables. We were attempting to move away from these.

As I understand it, the goal is to be able to find all selectors with a given filter value and remove it?  If so, can we maybe store the information on the selectr dom nodes itself maybe in a dataset attribute?

client/locations-list-map.js
```diff
@@ -1247,12 +1250,17 @@ function refreshList(data, filters) {
...
 }
 
-function onFilterChange(data, prefix, idx, selected, filters) {
+function onFilterChange(data, prefix, idx, selected, filters, selectr) {
   const primaryFilter = filters[prefix] && filters[prefix][Object.keys(filters[prefix])[idx]];
   if (!primaryFilter) {
     return;
   }
-
+  const temp = filters[prefix][Object.keys(filters[prefix])[idx]].name;
```
awong-dev, line 1258: nit: can we have a more descriptive variable name? I'm not 100% certain what the semantics of temp are...

client/locations-list-map.js
awong-dev: or maybe that's not possible...sorry, I don't think I understood your code.  I now wonder if this should just be an array of all Selectr objects. Those are basically constant for the duration of the page after initialization I think?

client/locations-list-map.js
```diff
@@ -1528,6 +1536,9 @@ function initMapSearch(data, filters) {
     e.preventDefault();
     resetMap(data, filters);
     $search.val('');
+    Object.keys(filterSelectors).forEach((current) => {
+      filterSelectors[current].clear();
```
awong-dev, line 1540: Hmm... so getting to this line, it looks like what we're trying to do is just call .clear() on all selectors. Is that accurate?

If so, do we actually need a map of the selector id to the selectr object? Could we get away with just an array of all selectr objects to clear?

client/locations-list-map.js
```diff
@@ -1528,6 +1536,9 @@ function initMapSearch(data, filters) {
     e.preventDefault();
     resetMap(data, filters);
     $search.val('');
+    Object.keys(filterSelectors).forEach((current) => {
```
awong-dev, line 1539: FYI, you can use Object.entries() to avoid the second lookup through the data structure.

client/locations-list-map.js
zhuochenwang: Hey Albert, thanks for the review! I kept it a map of filter id to selector objects in case someone deselects a selector. So the map only contains the currently selected ones, I can look for a way to achieve this with an array too.

client/locations-list-map.js
zhuochenwang: Yeah, I'm trying to give other methods like when we click on reset the map the access to the selected selector objects so that's why I declared it glob. I was trying to find a way but I couldn't find a way to go from the Dom element back to the selector objects that generated them. If that's possible, I think we can avoid the global variables.

client/locations-list-map.js
zhuochenwang: Gotcha, I think I steered away from the methods that did not have IE11 compatibility when writing the code.

awong-dev: ah yes! Good thinking!  Though in our code, we've been relying on babel and polyfills to handle this gap. Steve fixed some mistakes we had so it should be safe to use whatever.

zhuochenwang:
Needs review, currently storing reference to Selector objects in the DOM but according to the second comment on this stack overflow page (https://stackoverflow.com/questions/1402693/is-there-a-good-way-to-attach-javascript-objects-to-html-elements), it can lead to memory leakage.

awong-dev:
Let's go with it for now. The comment indicates it may leak. I would guess this happens in a refcount cycle if the javascript Object or closure holds the element and the element now refers to the Object.

However this is only a problem if we're removing the element from the dom. Since that's never happening, then when the Javascript "World" in Chrome is torn down along with the Document, all of it is dropped.

The real problem design wise is the code doesn't allow for us to register a map (eg ES6's WeakMap) without making it a module global.

If this annoys someone enough, they can refactor it later.

PR #1045, opened 2020-06-11
Clear out command links if missing row_id

awong-dev:


awong-dev:
@4kbt says this is spiritually correct which is good enough for me.

PR #1046, opened 2020-06-11
1023/language drop down in native language

zhuochenwang:
I didn't want to change a lot fo the code but this is functional from my testing. I left a comment in the code for setBananI18n.js about whether or not the entire for loop that was used to translate locale names is necessary.


client/locations-list-map.js
```diff
@@ -317,7 +317,16 @@ function multilineStringToNodes(input) {
 
 function createMakerMarkerContent(entry, separator) {
   // Text to go into InfoWindow
-  const contentTags = [ce('h5', separator ? 'separator' : null, ctn(entry.name))];
+  const contentTags = [];
```
nburt, line 320: it looks like this is code from @ngottlieb previous pr, do you know how that got in here?

middleware/setBananaI18n.js
```diff
@@ -61,11 +61,11 @@ module.exports = (req, res, next) => {
 
   const translatedLocales = [];
   let activeLocale;
-
+  // Do we still need the for loop?
```
nburt, line 64: we need to find the activeLocale still but don't need to translate it, we'd just need to also change the view that creates the locale dropdown

constants/locales.js
```diff
@@ -2,33 +2,41 @@ module.exports = [
...
   },
   {
     i18nString: 'ftm-italian',
     localeCode: 'it',
+    nativeLang: 'Italiano',
   },
   {
     i18nString: 'ftm-polish',
     localeCode: 'pl',
+    nativeLang: 'Polski',
```
nburt, line 25: Are we using i18nString anymore? If not, we can remove from here + en.json

zhuochenwang: I don't think we are, I kept just in case

client/locations-list-map.js
zhuochenwang: Ahh I got completely lost in version control and had some old commits. I tried to commit the newest copy of location-list.js but there must have been some mixup. I'll revert it to the deleted line.

middleware/setBananaI18n.js
zhuochenwang: Oh ok, so for loop stays to check for active locales but no more translation needed right? I'll change the code in a bit

client/locations-list-map.js
zhuochenwang: Yeah I checked the location-list.js and there are so many changes that I didn't intend to introduce. I only wanted to change setBanana and locale, do you know how I can fix this?

PR #1047, opened 2020-06-11
Language options from dropdown no longer translated but in their own respective language

zhuochenwang:
Changed locales.js, setBananaI18n.js, and en.json.

I still don't understand version control...


middleware/setBananaI18n.js
```diff
@@ -59,20 +59,12 @@ module.exports = (req, res, next) => {
     messages: config.messages,
   });
 
-  const translatedLocales = [];
   let activeLocale;
 
   locales.forEach((l) => {
```
nburt, line 64: rather than iterating through each locale you can use find so it stops when you find the active one:

`const activeLocale = locales.find((l)  => l.localeCode === locale)`

PR #1048, opened 2020-06-14
Do not put alert in the embed

awong-dev:


awong-dev:
Given the discussion from OSMS, I think maybe we should just remove the global alert from embeds? It doesn't seem to make as much sense if it's not the top-level page...

Thoughts?

Alternately, I can plumb through a parameter.

ngottlieb:
I support removing it from the embed, I think that makes sense. @linderr?

awong-dev:
@ngottlieb Your support is good enough for me! Let's merge it!

It's code...we can always revert.

linderr:
Thanks for working on this request @awong-dev  Content-wise I think we can prepare main page vs. shorter embed text in the future. A lot of our other collaborators actually really liked the alert when we workshopped it with them for feedback before shipping what @nburt coded --- they said it took work off their plate to notify their user base, and they were unclear how to address what was going on, on the headlines but welcomed our neutral-positive presentation.

awong-dev:
Okay. Putting it back is easy too. We'll just need to sort out some
conditional logic to disable it.

linderr:
@ngottlieb is always coming in clutch with conditional logic. Would be cool to see how Nick would approach this

ngottlieb:
hahah nothing too clever @linderr just the ol' if/else construct

tobiasdeml:
Yeah, disable or auto-shortening (i.e. saying all embeds will automatically take the first 20 words and then add elipsis and a "Read More" type link) would be a great solution to not take up nearly 50% of real estate of an embed :)

linderr:
@tobiasdeml the elipsis/read more option sounds like a good compromise! Would you like me to add you as a collaborator on our GitHub ? 😄

tobiasdeml:
Sure thing!

tobiasdeml:
Following up on this, it still takes up 50% plus of real estate on the embeds. Would be a high priority for us to get fixed.

PR #1056, opened 2020-06-15
844/Entry age filter

zhuochenwang:
For issue 844 raised by Charlie about entry ages. @stephmarisa and I think we got something functional for the entry age filter. Not completely ready for code review. Can someone help us decide whether or not we should use a less than logic (i.e. all entries within 1 week, within 2 weeks...) or a price range selector by Albert (i.e. entries with entry age between 1-7. 8-14...)?

### DIFF
--- a/client/locations-list-map.js
+++ b/client/locations-list-map.js
@@ -661,7 +661,6 @@ function getMarkers(data, appliedFilters, bounds, markerOptions) {
               if (entry.entry_age >= min && entry.entry_age <= parseInt(rangeArray[1], 10)) {
                 return true;
               }
-              return false;
             } else {
               if (entry.entry_age >= min) {
                 return true;
@@ -987,7 +986,6 @@ function getFlatFilteredEntries(data, filters) {
           if (entry.entry_age >= min && entry.entry_age <= parseInt(rangeArray[1], 10)) {
             return true;
           }
-          return false;
         } else {
           if (entry.entry_age >= min){
             return true;
--- a/client/locations-list-map.js
+++ b/client/locations-list-map.js
@@ -658,7 +658,7 @@ function getMarkers(data, appliedFilters, bounds, markerOptions) {
             const min = parseInt(rangeArray[0], 10);
             if (rangeArray.length === 2 && entry.entry_age >= min && entry.entry_age <= parseInt(rangeArray[1], 10)) {
               return true;
-            } 
+            }
             if (rangeArray.length === 1 && entry.entry_age >= min) {
               return true;
             }
@@ -980,7 +980,7 @@ function getFlatFilteredEntries(data, filters) {
         const min = parseInt(rangeArray[0], 10);
         if (rangeArray.length === 2 && entry.entry_age >= min && entry.entry_age <= parseInt(rangeArray[1], 10)) {
           return true;
-        } 
+        }
         if (rangeArray.length === 1 && entry.entry_age >= min) {
           return true;
         }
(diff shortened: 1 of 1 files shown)


client/locations-list-map.js
```diff
@@ -206,7 +206,11 @@ function createFilters(data) {
 
   const filters = {
     states: {},
+    entryAge:{},
   };
+  filters.entryAge.placeholder = "Entry age";
```
linderr: I think the public phrase could "Last Updated" here :)

linderr:
From our needs dataset, the vast majority will not be in the 1 week range. I still think integrating @4kbt's original as:
- this week
- two week ago (including 1 week entries)
- this month (including 1 week and 2 week entries)
- all (default)
makes sense because otherwise you won't see very many pins.

linderr:
Or, if you want to keep the ranges you've set, please allow multi-select support. Right now when you selected "2 weeks" "3 weeks" and "1 week" pills, the "1 week" filter is set to supercede/trump all other logic :)

linderr:
@stephmarisa @zhuochenwang do you guys want to try either the <7d, <14d, <30d **or** 1-7, 8-14, 15-21, 21+ price range logic with multiselect and see where you land? The first solution seems easier as a first past to get this up for people to use ASAP. The second solution can be a v2 on this filter menu and allow people more specific entry age control :)

client/locations-list-map.js
zhuochenwang: Gotcha, will change this when we go back and make our edits!

zhuochenwang:
@linderr I think we might go with the first options but change so it's this week, two weeks ago, and this month as Charlie suggested. Later on, we can work on the v2 of the logic with multi-select and price ranges.

linderr:
If your logic is <7, <14, <30 days you don't need multi-select because <14 would include <7 :)

zhuochenwang:
Sounds good, we will get on it later today/early tomorrow!

zhuochenwang:
Defining multiselect is kinda weird using Selectr and interferes with clearing filters and other things... So I might just try to implement v2 since that will be the final version we are aiming to publish.

ngottlieb:
Ah I get it now, @linderr was recommending the multi-select version as the "phase 2" which you just did as phase 1 instead. that's fine!

zhuochenwang:
@ngottlieb Hey nick, thanks for looking at the code. First time using the array.some() method which was interesting. I realized having to make one of the Selectr objects not multi-selectable might take more work for something we will eventually replace so I went ahead and did v2 of the filter. I also fixed the bug I mentioned on slack about the filters having an additional blank option that was due to the placeholder.

ngottlieb:
@zhuochenwang :+1: 🚢 

@linderr anything else on this you're wanting to see? Looks good to me. @4kbt may also have some input?

linderr:
Nice work @zhuochenwang & @stephmarisa ! 🎉 LGTM @ngottlieb 👍 
I'll ping @4kbt on Slack to let him know this is uppppp!! OSMS will also be satisfied, only a few more requests to go 😅

linderr:
Oops! I was too trigger happy --- just realized we missed adding an i18n tag for the filter and options :) @zhuochenwang @stephmarisa @ngottlieb

linderr:
@zhuochenwang @stephmarisa would you guys like to do the honours for posting the Issue to add onto your GitHub contributions count? Here's a visual example:
image

PR #1058, opened 2020-06-15
Made translations team blurb visible on Join Us page

NickB64:
The Translations team blurb on the Join Us page is no longer invisible.


linderr:
LGTM! You have unlocked d-none logic :)

Issue #1059, opened 2020-06-15
Join Us links switches back to US


NickB64:
When on the About Us page for countries other than the United States or languages other than English, clicking on the links for "Join Us", "translations team", and "internationalization team" (which all direct to the Join Us page) result in switching back to the United States for country and English as a language.

PR #1060, opened 2020-06-16
The three links leading to the Join Us page should now all retain the…

NickB64:
… current country/language the user is on


ngottlieb:
I actually have mixed feelings about this -- our current "Join Us" page is not i18n because we don't have the capacity to onboard someone who doesn't speak English. I could see an argument for changing that -- some of our leads in other countries could handle onboarding people in their language...but we haven't changed it yet.

@linderr any thoughts? 

If persisting locale becomes less important, then I would suggest just changing the link to `href="volunteer"` instead of the original `href="/volunteer"` which would accomplish half of the related issue here.

NickB64:
> I actually have mixed feelings about this -- our current "Join Us" page is not i18n because we don't have the capacity to onboard someone who doesn't speak English. I could see an argument for changing that -- some of our leads in other countries could handle onboarding people in their language...but we haven't changed it yet.
> 
> @linderr any thoughts?
> 
> If persisting locale becomes less important, then I would suggest just changing the link to `href="volunteer"` instead of the original `href="/volunteer"` which would accomplish half of the related issue here.

Hey Nick, that's a good point; I was thinking about that when changing the code. I was thinking, though, that it's possible a visitor would fancy visiting other sections of the site after they took one of these paths to visit the Join Us page, in which case their language and country would be inconveniently reset. It is true that the Join Us page is currently only depicted in English, so it might be confusing to the visitor if their language is not English, yet the content of the site is in English. Conversely, a visitor would probably be confused as to why their country (this might be an easy fix, like you said) and language changed back to the default after visiting the Join Us page. I guess it's just a tradeoff between the confusion caused by the language and content not aligning and the inconvenience of having to switch back to their country/language. I'm not a design artist by any stretch though, those are just my two cents :)

ngottlieb:
Good point! I was just trying to think of reasons to not embed JS logic into the translation JSON files hahah, but it does seem to be the most straightforward fix here.

ngottlieb:
resolve #1059

linderr:
Loving the Nick^2 convo here!

I agree with @ngottlieb that operationally, we don't have the capacity to actually receive folks who don't have some ability in English, unfortunately :( however, the jarring effect @NickB64 has identified & lack of language choice retention when trying to move onto a different subpage from Join Us is important.

Thanks for fixing and approving this PR, Nick^2!

linderr:
And easy way to address both of your concerns is just to put a short blurb at the top of Join Us that is flagged for translations via an i18n tag. Something like (can write a better one later):

"To stay fully coordinated, the common language across #findthemasks is English. However, we welcome anyone interested to join the community at Find the Masks! Reach out to us if you would like to help us become more international!" 

A fancier version of the fix could use the country `partials `logic? @NickB64 and the other interns have learned how they work now from helping us surface all of the country-specific About Us! 
On Join Us, instead of all languages leading to one generic 100% English-only page, we could only include the additional blurb on a single Join US-international page?

ngottlieb:
@linderr TBH I wasn't concerned about that, I was just trying to think of ways to avoid over-complicating our `**.json` with embedded Javascript (i.e. by suggesting we ignore half the problem @NickB64 was trying to solve instead), but what both you and he said makes it clear we shouldn't just ignore that. I have no problem with it just being an English page with no explanation but if you think we should add a small translatable blurb at the top, definitely make an issue for that and we'll get it done

linderr:
hahah @ngottlieb I'm trying to be cognizant of the sage advice you shared yesterday on the call about seeing the real issue/full context ;)

PR #1062, opened 2020-06-16
Revert "Do not put alert in the embed"

awong-dev:
Reverts findthemasks/findthemasks#1048


awong-dev:
@linderr @ngottlieb  Reverting.  Will try to insert an if statement based on parameter if possible tomorrow unless someone else gets to it first.

Issue #1053, opened 2020-06-14
Filter Performance Problems


ngottlieb:
Using filters on the US site has significant lag between user action and anything happen (same with reset, etc. -- any change to the filters). It is there on non-US pages too (proportionally to how many sites we have) but less noticeable. 

Solution should first attempt to resolve the performance issue by profiling the functions we're calling and seeing if we're doing anything silly.

If there are no optimizations we can make easily, the alternative is to dissociate the form control actions from the actual data filtering and add a loading spinner -- so, when you change a filter, that action is reflected immediately in the filter box and triggers a loading spinner while the filtering actually happens.

smattingly:
Some quick and dirty profiler work shows that this statement is taking all the time:
https://github.com/findthemasks/findthemasks/blob/27d01b77eb147f47e5a693daa7e6fa0dfdd519f0/client/locations-list-map.js#L980-L983

Maybe someone who knows more about Google Maps or our marker handling can see an easy solution? Caching some marker info instead of rebuilding it?

ngottlieb:
Interesting, maybe we can initialize the whole set of markers with the appropriate options and then only call `setOptions` on `secondaryMarkers`? Will have to look in more detail, but that's my first idea.

PR #1063, opened 2020-06-16
1051 Make Embed layout more dynamic and add toggle buttons on mobile …

eikhunter:
…and web

This uses the old default value of hideList if you want the list to be hidden by default but it can be open and closed regardless so the old defaults are removed and there will only be 1 option available to people who want to embed as you can just toggle both views.


views/index.handlebars
```diff
@@ -36,7 +36,7 @@
             <p class="i18n" data-i18n="[html]ftm-maker-request-near-me" style="margin-bottom: 0">
               Coordinate with a maker group near you. Questions? <a class="text-light" href="/faq#section-makers">Read our Maker FAQ.</a>
             </p>
-             <p class="i18n" data-i18n="[html]ftm-cta-link-back">
+            <p class="i18n" data-i18n="[html]ftm-cta-link-back">
```
eikhunter, line 39: I can change these defaults, as my refactor wants to add tabs where there are handlebar references

client/locations-list-map.js
```diff
@@ -1957,6 +1951,60 @@ const initGlobalAlert = () => {
...
+    mobileViewButtons[i].addEventListener('click', () => toggleMobileView(i));
+  }
+};
+
+const toggleEmbedList = () => {
+  const button = document.getElementById('js-ToggleListButton');
+  const container = document.getElementById('js-EmbedContainer');
+
+  if (container.classList.contains('list-open')) {
+    button.innerText = 'Open list';
```
nburt, line 1971: Can you add i18n for this? this is the wiki for it: https://github.com/findthemasks/findthemasks/wiki/Translation-Process

you'll need to add the strings to en.json, use $.i18n('string-key-name') in place of the hard coded strings, and then when you push up the PR the strings will be machine translated and a new PR will be opened against your current one that you can merge in

client/locations-list-map.js
```diff
@@ -1957,6 +1951,60 @@ const initGlobalAlert = () => {
   });
 };
 
+const initEmbedListEventListener = () => {
+  // Web toggle button
+  document.getElementById('js-ToggleListButton').addEventListener('click', toggleEmbedList);
```
nburt, line 1956: I get a JS exception on the homepage and /makers pages for this (works great on embed). @linderr do we want this button on the homepage too or just embeds? Either need to wrap this to check if it's an embed, or make it work for both

client/locations-list-map.js
```diff
@@ -1957,6 +1951,60 @@ const initGlobalAlert = () => {
...
+  // Set the web button text as well in case they have shrunk the screen
+  if (container.classList.contains('list-open')) {
+    webButton.innerText = 'Open list';
+  } else {
+    webButton.innerText = 'Close list';
+  }
+
+  for (let i = 0; i < mobileViewButtons.length; i++) {
+    if (index !== i) {
+      mobileViewButtons[i].dataset.active = 'false';
```
nburt, line 2001: any reason you're using strings here instead of booleans?

tobiasdeml:
Following up on this issue, would be great to have it implemented and would improve our usability of the embeds!

client/locations-list-map.js
eikhunter: I thought because of the data-active attribute ="true" it needs to equal a string or it throws an error, how else would you do this, I couldn't find anything on data-* could be a boolean value.

client/locations-list-map.js
eikhunter: @linderr I can implement this on the homepage too, wouldn't take a second just let me know

client/locations-list-map.js
eikhunter: I believe I've done it, worth getting another eye on it just in case, thanks Nate

nburt: nice looks good. I'm noticing the button appears before the map and list do, is it possible to render after map load or list load? Not worth holding up merge but would be nice to have

client/locations-list-map.js
nburt: ah I see, didn't quite get that this was added to the button as a data attribute, makes sense

client/locations-list-map.js
nburt: @eikhunter if we don't hear from Linda by am I think just only enable on the embed page and not on the homepage and then we can get this shipped. We can always follow up on adding it to the homepage

client/locations-list-map.js
eikhunter: I can't for the life of me figure this out as the `display: block` gets added once everything has loaded, but the element is hidden a couple of seconds before it, can't see anything at a glance but like you said not a merge breaker.

client/locations-list-map.js
eikhunter: I've just done this for the meantime, as I believe she has been moving house and we can take a separate look at the homepage

linderr: Great job, guys. Thanks for being considerate but also working around me! Settled and back on it today. For homepage, I think it's the right direction to head towards to support v2 mobile UX/UI design that is toggleable between list and map to make mobile use friendlier and avoid you guys having to maintain 2 code bases between desktop and mobile. However, the homepage currently has the default Google Maps full screen map view toggle. This could be redundant if we implement @eikhunter's toggle and should be removed. Will create issues

client/locations-list-map.js
linderr: There are known performance issues of the list/map - they load increasingly slowly, especially on the larger US dataset. @ngottlieb has an issue for it on #1053 if anyone wants to investigate further :) thanks guys!

PR #1066, opened 2020-06-17
Added i18n tag to entryage filters, need smartling

zhuochenwang:
Added the i18n tags to the entry age filters. Since there are only 5 items to be translated every time I didn't write a for loop but let me know if I should. Need to make a PR so that the Smartling bot can translated the new entries to the other languages.


client/locations-list-map.js
```diff
@@ -208,11 +208,11 @@ function createFilters(data) {
     states: {},
     entryAge: {},
   };
-  filters.entryAge['1-7'] = { name: '1-7 days ago', isSet: false, value: '1-7' };
-  filters.entryAge['8-14'] = { name: '8-14 days ago', isSet: false, value: '8-14' };
-  filters.entryAge['15-21'] = { name: '15-21 days ago', isSet: false, value: '15-21' };
-  filters.entryAge['21+'] = { name: '21+ days ago', isSet: false, value: '21' };
-  filters.entryAge.placeholder = 'Last Updated';
+  filters.entryAge['1-7'] = { name: $.i18n('ftm-entry-age-1-7'), isSet: false, value: '1-7' };
```
nburt, line 211: I'll merge in smartling translations to the pr, agree a for loop isn't really necessary but maybe just declare as one object?

```
entryAge: {
  '1-7': { name: $.i18n('ftm-entry-age-1-7'), isSet: false, value: '1-7' },
  ...
}
```

zhuochenwang:
Ready for final review. The translations seemed all good from my testing. Changed the format of the entry age filter in the code to be one object as @nburt suggested! Let me know if there is anything else I can change.

awong-dev:
looks good to me! merging.

PR #1070, opened 2020-06-18
1069/documentation part one

zhuochenwang:
Made some simple comments to test out how I should help document location-list. Would really appreciate any feedback! Also, should I keep one branch for the documentation stuff?


client/locations-list-map.js
```diff
@@ -2051,6 +2054,8 @@ $(() => {
   initContactModal();
   initGlobalAlert();
 
+  // Get the relevant json data file based on country and requester/maker from findthemasks.com and
```
nburt, line 2057: we refer to "requester/maker" as a "dataset" throughout the codebase. I think that would be good to call out.

zhuochenwang: Gotcha, I'll make that change on the next commit!

PR #1071, opened 2020-06-20
Add "notices=none" into /embed to disable notices.

awong-dev:


applicationRoutes.js
```diff
@@ -55,6 +55,10 @@ router.get(['/give', '/give.html', '/embed'], (req, res) => {
     }
   }
 
+  // Get notice configuration. "none" is the only support value currently
+  // and disables all notices.
+  res.locals.notices = req.query['notices'];
```
nburt, line 60: if we're only support one value ("none") should this just be a boolean? `res.locals.hideNotices = req.query['notices'] === 'none'`;

awong-dev: figured we'd pipe the logic through in case we wanted more values over time...

PR #1074, opened 2020-06-22
1069/documentation part two

zhuochenwang:
Second part of the documentation which focuses on the initMap() portion of the location-list.js. A lot of the methods that are called already have some description, so I just added little changes/supplements here and there. Let me know if there is an area of the code that can be worked on/documented next!

### DIFF
--- a/client/locations-list-map.js
+++ b/client/locations-list-map.js
@@ -1860,7 +1860,7 @@ function initMap(data, filters) {
   });
 
   const mapBounds = gMap.getBounds();
-  
+
   // On initialization, set global variable to current center of map
   if (mapBounds) {
     const mapCenter = mapBounds.getCenter();
@@ -1878,7 +1878,7 @@ function initMap(data, filters) {
   loadOtherCountries();
 
   // Add map control for adding additional datasets
-  // Checking additional layers in the google maps legend will fetch and process the related data (if it does not exist 
+  // Checking additional layers in the google maps legend will fetch and process the related data (if it does not exist
   // in datasetData) and push it to the secondary marker set for it to be visible.
   const onChange = (dataset, event) => {
     const { checked } = event.target;
--- a/client/locations-list-map.js
+++ b/client/locations-list-map.js
@@ -46,7 +46,7 @@ const gOtherMarkers = [];
 let gPrimaryCluster = null;
 // Secondary + other markers shown in secondary cluster
 let gSecondaryCluster = null;
-// The longitude and latitude of the center of Google Maps
+// The longitude and latitude of the center of Google Maps.
 let gCurrentViewportCenter = {};
 
 const gDatasetMarkers = {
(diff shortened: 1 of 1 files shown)


client/locations-list-map.js
```diff
@@ -46,7 +46,7 @@ const gOtherMarkers = [];
 let gPrimaryCluster = null;
 // Secondary + other markers shown in secondary cluster
 let gSecondaryCluster = null;
-
+// The longitude and latitude of the center of Google Maps
```
awong-dev, line 49: Period at end of sentence please. :)

client/locations-list-map.js
```diff
@@ -64,6 +64,7 @@ const gDatasetMarkers = {
   },
 };
 
+// All other datasets that are not the dataset we are currently on
```
awong-dev, line 67: Period at end of sentence please. :)

Though this makes me think the variable is renamed. Should it be OTHER_DATASETS?

If that can make it clear enough to not need the comment...that'd even better.

client/locations-list-map.js
```diff
@@ -231,10 +232,12 @@ function createFilters(data) {
     },
   };
 
+  // All states are included as potential filters through query and not selectable.
```
awong-dev, line 235: "through query and not selectable." <-- this is slightly unclear to me. Are you saying

States are always included as potential filters and clients select it via the URL query parameters
even if they are not selectable?

Maybe?

client/locations-list-map.js
```diff
@@ -231,10 +232,12 @@ function createFilters(data) {
...
   for (const state of Object.keys(data)) {
     filters.states[state] = { name: state, isSet: false };
   }
 
+  // Load all other possible whitelisted filters with their accepted, translated values
```
awong-dev, line 240: whitelisted -> allowlisted

As an industry we should prefer allowlist/blocklist over whitelist/blacklist going forward.

client/locations-list-map.js
```diff
@@ -1527,6 +1530,7 @@ function getDatasetFilename(dataset, countryCode) {
   return `/${dataset}-${countryCode}.json`;
 }
 
+// Grab the country codes of other countries and load their json data too
```
awong-dev, line 1533: Period. :)

Also, it's preferable to explain the side effects of a function.  (Honestly, it's preferable to not even have global side effects, but alas...too big of a refactor).

client/locations-list-map.js
```diff
@@ -1854,6 +1861,7 @@ function initMap(data, filters) {
 
   const mapBounds = gMap.getBounds();
 
+  // On initialization, set global variable to current center of map
```
awong-dev, line 1864: Specify which variable.

client/locations-list-map.js
```diff
@@ -1921,6 +1931,7 @@ function initMap(data, filters) {
 
   const legend = generateGMapDatasetLegend(onChange);
 
+  // Append the newly created legend with its event listener to top left of google Maps
```
awong-dev, line 1934: .

client/locations-list-map.js
```diff
@@ -1870,6 +1878,8 @@ function initMap(data, filters) {
   loadOtherCountries();
 
   // Add map control for adding additional datasets
```
awong-dev, line 1880: Period.

client/locations-list-map.js
zhuochenwang: Yeah its always filtering out the current dataset so I wonder if I can just rename it.

awong-dev: please do! Better variable names >> comments.

PR #1077, opened 2020-06-24
Use timestamp for entry_age filter

zhuochenwang:
Changing the way we are filtering by entry_age to use the timestamp instead of the entry_age value. I also changed the filter ranges so that the most recent entries that are updated 0 days ago are included for the filter updated 1-7 days ago.

### DIFF
--- a/client/locations-list-map.js
+++ b/client/locations-list-map.js
@@ -681,7 +681,7 @@ function getMarkers(data, appliedFilters, bounds, markerOptions) {
             if (rangeArray.length === 2) {
               const maxDate = new Date();
               maxDate.setDate(maxDate.getDate() - parseInt(rangeArray[1], 10));
-              if (entryDate < maxDate){
+              if (entryDate < maxDate) {
                 return false;
               }
             }
@@ -1009,7 +1009,7 @@ function getFlatFilteredEntries(data, filters) {
         if (rangeArray.length === 2) {
           const maxDate = new Date();
           maxDate.setDate(maxDate.getDate() - parseInt(rangeArray[1], 10));
-          if (entryDate < maxDate){
+          if (entryDate < maxDate) {
             return false;
           }
         }
--- a/client/locations-list-map.js
+++ b/client/locations-list-map.js
@@ -207,7 +207,7 @@ function createFilters(data) {
   const filters = {
     states: {},
     entryAge: {
-      '1-7': {
+      '0-7': {
         name: $.i18n('ftm-entry-age-1-7'),
         isSet: false,
         value: '1-7',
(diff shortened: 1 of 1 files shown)


client/locations-list-map.js
```diff
@@ -670,13 +670,17 @@ function getMarkers(data, appliedFilters, bounds, markerOptions) {
           }
         });
         if (hasEntryFilter) {
+          const today = new Date();
+          const dateConversion = 86400000;
```
awong-dev, line 674: hmm... manual date manipulations are always a little tricky.  For example, how does this interact with leap-years or daylight savings? I tend to think of "dates" as two concepts: calendar dates for human consumptions, and timestamps for calculating deltas between intervals.   In the former, you never do it by hand with math if you can avoid it. In the latter, you ensure everything is in utc or something and as best as possible, convert to seconds (like you did with getTime() to calculate...but even then, there are funny things with leap seconds where dates are no longer monotonic...

Anyways, that's a long way of saying, I believe we re trying to work in calendar dates here, not timestamps.

Is it possible to use inequality operators with the appropriate closed/open intervals?  (note, Javascript date API is broken on equality...cause javascript sucks..:  https://stackoverflow.com/questions/492994/compare-two-dates-with-javascript)

And then use the calendar date functions like https://stackoverflow.com/questions/1296358/subtract-days-from-a-date-in-javascript to generate your interval bounds?

client/locations-list-map.js
```diff
@@ -672,14 +672,20 @@ function getMarkers(data, appliedFilters, bounds, markerOptions) {
         if (hasEntryFilter) {
           if (!Object.keys(entryAge).some((entryFilter) => {
             const rangeArray = entryFilter.split('-');
-            const min = parseInt(rangeArray[0], 10);
-            if (rangeArray.length === 2 && entry.entry_age >= min && entry.entry_age <= parseInt(rangeArray[1], 10)) {
-              return true;
+            const entryDate = new Date(entry.timestamp);
+            const minDate = new Date();
+            minDate.setDate(minDate.getDate() - parseInt(rangeArray[0], 10));
```
awong-dev, line 677: This is okay for the current codebase, but we're getting a little close to using a human-readable string as a computer-parseable value. That tends to be more fragile than, say, having a map of the exact value to a number because it locks the string format into a structure in a way that has no statically verifable or searchable indication.

client/locations-list-map.js
```diff
@@ -994,14 +1000,20 @@ function getFlatFilteredEntries(data, filters) {
     if (entryAge) {
       if (!Object.keys(entryAge).some((entryFilter) => {
         const rangeArray = entryFilter.split('-');
-        const min = parseInt(rangeArray[0], 10);
-        if (rangeArray.length === 2 && entry.entry_age >= min && entry.entry_age <= parseInt(rangeArray[1], 10)) {
-          return true;
+        const entryDate = new Date(entry.timestamp);
```
awong-dev, line 1003: This code looks duplicated. Can we pull out a function?

PR #1088, opened 2020-07-01
Update filter now also changes secondary datasets

zhuochenwang:
To address the issue raised by @tobiasdeml, I just added some code to showMarkers() in location-list to check for additional datasets like makers. This should work for any additional datasets we want to potentially add in the future too!


tobiasdeml:
Thanks so much!

PR #1089, opened 2020-07-01
1081/unit test cloud functions

zhuochenwang:
Added jest to dev dependencies. Edited travis.yml so that it runs npm run test which would execute a simple jest test of whether 1+2=3.


functions/unit_Test/sum.js
```diff
@@ -0,0 +1,4 @@
+function sum(a, b) {
+    return a + b;
+  }
+  module.exports = sum;
```
awong-dev, line 4: indentation looks odd...

Also, why the weird capitalization on the directory unit_Test? Since this is just a smoke-test of our unittest framework, there's no need to create a separate module for a tests module. Just dump it inside the .test.js file.

package.json
```diff
@@ -17,7 +17,8 @@
     "lint-client": "eslint 'client/**/*.js' --config 'client/.eslintrc.json'",
     "lint-node": "eslint '**/*.js' --ignore-pattern 'client/**/*'",
     "start": "node server.js",
-    "webpack": "NODE_ENV='production' webpack -p --config webpack.config.js"
+    "webpack": "NODE_ENV='production' webpack -p --config webpack.config.js",
+    "test": "jest"
```
awong-dev, line 21: Will this accidentally run the unittests in functions/ too?

zhuochenwang:
Hey, @awong-dev @nburt, I wrote some of the tests that you mentioned on slack for testing rootRoutes.js. However, I'm a little confused about whether I did it right. Right now, I just copied over the sendDataJsonFromCache() method and tested some of the functionalities through mocking sendDataJson and the https interaction but I'm not sure if it's "right". Was wondering if you guys could take a look when you have time and give me some feedback!

PR #1094, opened 2020-07-06
1090/root routes unit test

zhuochenwang:
Not the final product, but I got one small unit test done for rootRoutes.js to test whether or not caching is working. I had to make sendDataJson() a separate file and import it so I can mock its behavior. Also added two new commands to travis to test different unit tests in root (excluding functions) and /functions.


package.json
```diff
@@ -24,7 +23,8 @@
     "lint-node": "eslint '**/*.js' --ignore-pattern 'client/**/*'",
     "start": "node server.js",
     "webpack": "NODE_ENV='production' webpack -p --config webpack.config.js",
-    "test": "jest"
+    "test-root": "jest --testPathIgnorePatterns='/functions'",
```
awong-dev, line 26: Hmm... so this is conceptually incorrect as it is using the root-package "jest" installation to run the tests in the /function package which breaks the hermeticity.

travis should instead do npm install and npm run while inside the functions directory... thinking about this more, we might want to do in an independent CL.

nburt: I don't have a ton of experience with travis but can we have a different config file for it in the functions dir?

Maybe for now the "test-root" can be renamed to "test" and we just remove the "test-functions" script? That way we can get this change in and then figure out how to work in testing for the functions dir

zhuochenwang: @awong-dev Do you think I should leave the /functions folder as is and revert it back to the old jest test? As far as I know, no one is currently working on the cloud function unit tests just yet. We could get to to when we are working on them later on.

awong-dev:
Leave it untested for now..

zhuochenwang:
Hey @nburt @ngottlieb, I think I was able to mock the https request part using nock as Nick suggested, but I'm a little stuck on testing the asynchronous callback. I'm trying to use done() but it seems to be throwing errors and triggering the expect tests to begin at weird times. Would you guys taking a look at what I have right now and see if there is any way I can approach the issue? I moved the callback function to a different module under sendDataJson.js (name of that file still TBD) so I can mock and spy on the functions when necessary.

sendDataJsonFromCache.test.js
```diff
@@ -0,0 +1,48 @@
...
+test('Testing whether it can insert a new cache entry', done => {
+  const spyCallBackFn = jest.spyOn(mockDataJson, 'updateCachedData');
+  const today = new Date();
+  const cache = {
+    US: {
+      expires_at: today.setDate(today.getDate() - 1),
+    },
+  };
+
+  const countryCode = 'US';
```
awong-dev, line 33: Ah...this is interesting.  Here is a classic mock vs fake test + code design situation.

The function sendDataJsonFromCache() does quite a few things:
  (a) Implicitly makes some assumptions about the structure of the input variables (booo dynamic languages)
  (b) Checks `cache` to see if an entry is expired
  (c) If it is expired, makes an http request to a server
       (c.1) generates a specific path pattern to the server based on the inputs
       (c.2) accumulates the response body into some full byte buffer
       (c.3) Sets up an error handler for network errors.
       (c.4) On happy path, writes result into cache
       (c.5) Calls send function.
  (d) On a cache hit, good response, or bad response, write data back from the cache to the user setting proper http status codes.

This is a lot of territory to cover, so "Testing whether it can insert a new cache entry" ends up being closer to an end-to-end test than a unittest where you have to setup some version of most of the system (eg using a complex nock setup to introduce "behavior" to the mock) to make the large assertion.

If you break it down....

Some of these, such as (a) and (c.2) are possibly too low-value to be truly worth "testing". And expected behavior might just accidentally get asserted by your other tests.

Some of these, such as  (b), and (c.1), are clearly self-contained chunks of business logic. This is classic unit-test fodder and writing a unittest for these is basically pulling out a function can controlling the inputs to call them like you've been doing.

Some of these, such as (c.3) and (c.5), are very clearly just "integrations" between "units" of functionality. In these, though we're using a "unit test" framework, we're actually writing an integration test, which means the test should, ideally, just assert that the right function was called X times, and that the calling code handle Foo/Bar/Baz results and exceptions when it was called. In particular, an ideally written test would function correctly without needing to accidentally encode *how* the function processes input to output.  So the `.get()` for mockNode explicitly specifying the format of the string is an example of making an implicit assertion (that the path construction is correct...which should have been covered by a simple unittest of c.1). This leads to over-testing because now 2 test break for the exact same reason. It also means you have to emulate the full API of https to get the test which, as you noticed, gets complicated.

Some of these, like (d) still have too much behavior in the concept. We want to assert that sendDataJson is always called no matter what errors are returned.  We don't want to, in the same test, assert that sendDataJson produces 200s or 404 or the test gets too big.

So taking a step back, my though is the test case topic itself is not specific enough.  You want something like:

(1) "Expired Cache Entry issues https request" and all you want to test is that the code path ends up leading to something that makes an call with roughly the right parameters. No need to even worry if the right data is passed back.

(2) Another test like "Successful https fetch updates cache" where you just return a static well formatted data blob to the function and check that `cache` is updated appropriately.

etc. etc...

Now in (1) and (2), you have a very very narrow assertion per test. You can then restructure code to make that assertion possible. Currently the function is `sendDataJsonFromCache(cache, prefix, countryCode, res)`. In (1), you only want to ensure that you end up constructing the correct HTTPS request.  There are 2 ways to do this.  (1) you could use nock to test it directly. Or two, you extract the parts into an internal function that you can "dependency inject."  With nock, cut the dependency lower with less abstraction, but it means you have a wider API to emulate (adding complexity to the test).  With an internal function, you add more abstraction complexity, but the "mock" surface can be much smaller.

Example of internal function would be

```
function getDataFromServer(cache, prefix, countryCode) {
   ... Do all the https stuff here and returns the data + error code...this function might need to return a Promise() ...
}

function writeCachedDataToResponse(cache, prefix, countryCode, res) {
}

function sendDataJsonFromCache(cache, prefix, countryCode, res) {
   return sendDataJsonFromCacheInternal(cache, prefix, countryCode, getDataFromServer, writeCachedDataToResponse);
}

sendDataJsonFromCacheInternal(cache, prefix, countryCode, res) {
  if (cache is expired) {
    const new_data = await getDataFromServer(cache, prefix, countryCode);
    ... put in cache...
  }

  writeCachedDataToResponse(cache, prefix, countryCode, res);
}
```

And you can imagine what `sendDataJsonFromCacheInternal()` looks like.  Now you only need to generate a mock for `getDataFromServer()` to assert 1 and you can do that w/o worrying at all about https, etc which leads to a clean dependency cut via dependency injection.

...this example is written using explicitly dependency injection (which I prefer a bit, but I'm oldskool).  With JS and Jest, I think you can use Jest.mock() to just replace the `getDataFromServer` directly allowing you to merge sendDataJsonFromCacheInternal() and sendDataJsonFromCache() without the explicit dependency injection.

...I think you were already half on this path, but the functionality cut boundaries in the current iteration are leading you to need to emulate too much of the https API behavior in order to "assert" the business logic that we care about.

... of course, all this is because I wrote a spaghetti mess to begin with. Sorry!

package.json
awong-dev: It's fine what you have for now. We can fix it later.

.travis.yml
```diff
@@ -4,4 +4,5 @@ node_js:
 
 script:
   - npm run lint
-  - npm run test
+  - npm run test-root
```
awong-dev, line 7: Just rename this to test and don't worry about test-functions yet.

This setup isn't quite right as you are still using the node_modules installed at root to run the tests for things in /functions.

But let's get what we have in first.

rootRoutes.js
```diff
@@ -6,66 +6,36 @@ const { createProxyMiddleware } = require('http-proxy-middleware');
 const applicationRoutes = require('./applicationRoutes');
 const apiRoutes = require('./apiRoutes.js');
 const countries = require('./constants/countries.js');
+const methods = require('./sendDataJson');
```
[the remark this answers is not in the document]
awong-dev: This comment doesn't seem addressed?  I would have expected not to see `methods` anywhere in this file.

FYI, usually it's clearer to add a reply to each review comment, even if it's just "ack".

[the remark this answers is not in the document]
zhuochenwang: Wait I completely missed this comment, my bad! I created the object instead of using destructing is because we need the object for jest.spyOn(object, 'method name') assignment!

.travis.yml
zhuochenwang: Gotcha, since the next thing Nick and I are working on are unit tests for the cloud functions, we will look more into that

PR #1099, opened 2020-07-27
unit tests for toDataByLocation first draft

stephmarisa:
added some unit tests for toDataByLocation. Any feedback would be appreciated!

### DIFF
--- a/client/toDataByLocation.js
+++ b/client/toDataByLocation.js
@@ -34,4 +34,4 @@ module.exports = function toDataByLocation(data, datasetKey) {
   });
 
   return dataByLocation;
-}
+};
--- a/client/toDataByLocation.js
+++ b/client/toDataByLocation.js
@@ -1,4 +1,4 @@
-export default function toDataByLocation(data, datasetKey) {
+module.exports = function toDataByLocation(data, datasetKey) {
   const headers = data.values[1];
   const approvedIndex = headers.indexOf('approved');
   const stateIndex = headers.indexOf('state');
--- a/client/toDataByLocation.test.js
+++ b/client/toDataByLocation.test.js
@@ -1,10 +1,10 @@
-const toDataByLocation = require('./toDataByLocation');
+const toDataByLocation = require('./toDataByLocation.js');
 
 test('filters out entries without approval', () => {
   const fakeDataApproval = {
     values: [
       [],
-      ['approved', 'city', 'state', 'lat', 'lng'], // headers 
+      ['approved', 'city', 'state', 'lat', 'lng'], // headers
       ['x', 'Seattle', 'WA', '47.6674625', '-122.3795306'], // valid entry
       ['', 'Burien', 'CA', '57.4585642', '-100.3339504'], // invalid approval
     ],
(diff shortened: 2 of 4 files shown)


client/toDataByLocation.js
```diff
@@ -1,4 +1,4 @@
-export default function toDataByLocation(data, datasetKey) {
+module.exports = function toDataByLocation(data, datasetKey) {
```
nburt, line 1: do you know why you needed to change this? Does our jest setup include babel? https://jestjs.io/docs/en/getting-started#using-babel

client/toDataByLocation.test.js
```diff
@@ -0,0 +1,87 @@
...
+        values: [
+            [],
+            ["approved", "city", "state", "lat", "lng"], // headers 
+            ["x", "Seattle", "WA", "47.6674625", "-122.3795306"], // valid entry
+            ["", "Burien", "CA", "57.4585642", "-100.3339504"], // invalid approval
+            ["x", "Seattle", "WA", "", "-122.3795306"], // invalid latitude 
+            ["x", "Seattle", "MO", "30", ""], // invalid longitude
+        ]
+    };    
+    Object.values(toDataByLocation(fakeDataApproval,'makers')).forEach((stateObj) => {
```
nburt, line 14: the nested forEachs can be little difficult to read. I generally do something along the lines of:

```
const expected = {
  makers: {
    CA: {
      cities: [
        { name: 'Burien', entries: [ { lat: "57.4585642", lng: "-100.3339504", approved: 'x' } ] }
      ]
    }
  }
}

expect(toDataByLocation(fakeDataApproval, 'makers')).toEqual(expected);
```

(ignore that the example I gave was burien which shouldn't be in there 🤦) I like this style since it's really explicit about what you're expecting back.

client/toDataByLocation.test.js
```diff
@@ -0,0 +1,87 @@
+const toDataByLocation = require('./toDataByLocation');
+
+test('Testing that an entry wont go through without approval or valid lat and lng', () => {
```
nburt, line 3: nit: more stylistic but I'd remove the "testing that" piece and just describe the test. we already know that this is a test so that's just noise. Something like `test('filters out entries without approval or a valid lat and long', () => ...)`. To break things down even further, I might even have separate tests for filtering out based off the approval vs. invalid lat / lng

client/toDataByLocation.test.js
stephmarisa: Oh I see.. that makes a lot of sense. I wasn't sure if I should mimic the structure of the expected output or pass in some fake data and check the output is right (what i was doing previously).

stephmarisa: that makes a lot of sense! Also, just checking, wouldn't the structure of the expected  output be
`  const expected = {
    "WA": {
      "cities": {
        "seattle": {
          "entries": [
            {"approved": "x", "city": "Seattle", "datasetKey": "makers", "lat": "47.6674625", "lng": "-122.3795306", "state": "WA"}
          ]
        }
      }
    }
  }`

nburt: yes, totally. Sorry was just guessing off the top of my head!

client/toDataByLocation.js
nburt: just wanna follow up here, do you know why this was necessary?

PR #1101, opened 2020-07-30
Wrote basic unit tests for parseAirtableData()

zhuochenwang:
This is the final method that @awong-dev suggested we play around with unit tests! Since Albert has always given us descriptions of the method and what to test for the previous 3 PRs, we wanted to challenge ourselves and do the same for this one. Here's what we have so far!

The method has 2 main functionality:

1. Map all the columns from the Airtable data to a new object, attaching new ftm-names whenever we match the regex in addition to specific value translators.

2.  Go through each row of the data and generate a useful data structure from it.

What we want to test:

For 1, we want to make sure that all columns are mapped to the new object. For columns that matches the regex, we create a new property called ftm_name with the matched string. Finally, we want to make sure all columns in the new object has a default value translator function with specific column types having a specific value translator.

For 2, we want to make sure that a new entry in the data object is generated for each row and that the key-value pairs for these entries contain the column name or ftm_name mapped to the translated value.

We think we were able to achieve most of the tests with a problem in checking the specific value translator functions since we can't check equality between anonymous functions inside the object. Please let us know what we should change/improve upon!

### DIFF
--- a/functions/airtable-shared-view.js
+++ b/functions/airtable-shared-view.js
@@ -39,11 +39,8 @@ function createSelectTranslator(colMetadata) {
 
 function identityTranslator(x) { return x.trim(); }
 
-// Takes the airtable json from readSharedViewData and converts it into an array
-// of values in the style FindTheMasks expects.
-function parseAirtableData(airtableData) {
-  // Get all the column names and data translation setup.
-  //
+// Get all the column names and data translation setup.
+function mapColumns(airtableDataColumns) {
   // Column types are:
   // "type": "foreignKey",
   // "type": "formula",
@@ -53,7 +50,7 @@ function parseAirtableData(airtableData) {
   // "type": "select",
   // "type": "text",
   const columnNames = {};
-  for (const col of airtableData.data.table.columns) {
+  for (const col of airtableDataColumns) {
     const match = col.name.match(/\[(\S+)\]$/);
     if (match) {
       col.ftm_name = match[1];
@@ -70,6 +67,13 @@ function parseAirtableData(airtableData) {
 
     columnNames[col.id] = col;
   }
+  return columnNames;
+}
+
+// Takes the airtable json from readSharedViewData and converts it into an array
+// of values in the style FindTheMasks expects.
+function parseAirtableData(airtableData) {
+  const columnNames = mapColumns(airtableData.data.table.columns);
 
   // Go through data rows and create a data struct for us.
   const data = [];
@@ -95,4 +99,11 @@ function parseAirtableData(airtableData) {
   return data;
 }
 
-module.exports = { readAirtableSharedView, parseAirtableData };
+const unitTestFunctions = {
+  mapColumns, 
+  createMultiSelectTranslator,
+  createSelectTranslator,
+  identityTranslator,
+}
+
+module.exports = { readAirtableSharedView, unitTestFunctions, parseAirtableData };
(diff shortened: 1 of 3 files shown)


functions/package.json
```diff
@@ -8,7 +8,7 @@
     "start": "npm run shell",
     "deploy": "firebase deploy --only functions",
     "logs": "firebase functions:log",
-    "test": "jest"
+    "test": "jest --verbose"
```
nburt, line 11: I might take this out just to not clog up logs for travis ci / etc. You can also pass args while running locally: e.g. `npm run test --verbose`

functions/unittest/parseAirtable.test.js
```diff
@@ -0,0 +1,70 @@
...
+      id: 'first',
+    },
+    {
+      name: 'Nothing should match here',
+      type: 'select',
+      id: 'second',
+    },
+  ];  
+  test('If the column name has a string encapsulated in [], regex should match and a ftm-name property should be made', () => {
+    const copyTestColumn = JSON.parse(JSON.stringify(testColumn));
```
nburt, line 18: might need babel to do this but another way to duplicate an object is:

```
const copyTestColumn = { ...testColumn }
```

Without babel I think this would work:
```
const copyTestColumn = Object.assign({}, testColumn);
```

parsing + stringifying works but it's a little less wonky

functions/airtable-shared-view.js
```diff
@@ -95,4 +99,11 @@ function parseAirtableData(airtableData) {
   return data;
 }
 
-module.exports = { readAirtableSharedView, parseAirtableData };
+const unitTestFunctions = {
```
nburt, line 102: I might just export the functions you wanna test rather than doing so under a "unitTestFunctions" and then test the translators separately.

functions/unittest/parseAirtable.test.js
```diff
@@ -0,0 +1,70 @@
...
+            name: 'Nothing should match here',
+            type: 'text',
+            id: 'second',
+          },
+        ],
+        rows: [
+          {
+            id: 1,
+            cellValuesByColumnId: {
+              first: ' this string has an intial space to be trimmed by default valueTranslator',
```
nburt, line 53: this test seems to be testing a few different things
- that the data structure is created correctly
- that the valueTranslator trims whitespace

I think that you could split this up to test the parseAirtableData function is returning the correct structure and then have different tests for the logic within each translator

functions/package.json
zhuochenwang: Oh I did not know that! I'll change on the next commit!

functions/unittest/parseAirtable.test.js
zhuochenwang: Nice! I only did a brief search about cloning JS objects and the parsing+stringifying method was the first one that showed up so I just went with this. I'm not too sure if we installed the babel-jest extension but I'll try the Object.asisgn()

functions/airtable-shared-view.js
zhuochenwang: Sounds good, I don't even remember why I tried to make a separate object for this lol

functions/unittest/parseAirtable.test.js
zhuochenwang: Ok, I did not actually go ahead and write tests about the functionality for the different valueTranslators() because I wasn't sure if they are up for change or not. I'll separate the test for the correct data structure and create some new tests for the functionalities of the valueTranslator().

